Issue #10, opened 2025-10-28
Feature: Run from node script


tkadlec:
Right now, the agent has to be run via a CLI. We should make it possible to run from within a node script as well.

sergeychernyshev:
PR submitted by @dangayle for this work is merged in 65db113ea9b493a343179ca52657b7bea3fe29cc

PR #26, opened 2025-11-02
Add programmatic API support (launchTest)

dangayle:
Introduces a programmatic API for running tests from Node.js scripts while maintaining backward compatibility with the CLI.

Changes:
- Extract core test execution into executeTest() function
- Add launchTest() as public programmatic entry point 
- Normalize options via normalizeConfig() for both CLI and programmatic use
- Make temporary context directories unique per test to prevent profile lock conflicts
- Add parseOption() helper to handle both string and object option values
- Update package.json main field to point to index.js
- Add comprehensive tests for programmatic API
- Add artifact structure test comparing CLI and API output
- Update README with programmatic usage example

Implements #10

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,12 +5,15 @@ import { TestRunner } from './lib/testRunner.js';
 import { ChromeRunner } from './lib/chromeRunner.js';
 import { log } from './lib/helpers.js';
 import { normalizeConfig } from './lib/config.js';
+import { DEFAULT_OPTIONS } from './lib/defaultOptions.js';
 
 /**
  * Execute a test with raw options.
+ * Internal function that handles the core test execution flow.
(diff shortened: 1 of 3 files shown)


sergeychernyshev:
Thank you for this detailed PR, @dangayle and a related #33 @43081j - will take a deeper look at them this week.

For this to filly work, we'll also need to publish the npm module, which will require some infra configurations on our end, obviously as well.

sergeychernyshev (approved):
Looks great, will be merging shortly.

sergeychernyshev:
Was merged in 65db113ea9b493a343179ca52657b7bea3fe29cc

Thank you for submitting this, @dangayle!

PR #33, opened 2025-11-05
feat: only normalize options for CLI

43081j:
This exports `launch` - a function which takes `LaunchOptions` and runs
the browser the same way the CLI does.

The CLI has then moved to `lib/cli.js` and calls `launch` internally
with normalized options.

This means the various CLI arg normalizations (e.g. parsing cookie
objects) happen only in the CLI, while the underlying code expects a
strong type.


sergeychernyshev:
@43081j I merged @dangayle's PR and added a couple of your fixes.

Can you take another look and see what else is missing?

Re type definitions, we are hoping to move the code to TypeScript in the near future as well so it will be a bit more strict and provide good typing and code completion for programmatic execution.

43081j:
I still don't think it's right that the launch function normalises options. If this was in typescript already, it would be a lot clearer why. 

The programmatic usage should accept a strong type, but the cli can't do that since all parameters are strings or booleans. I don't think we should be letting the cli normalisation leak into the underlying API 

By doing so, it means everything is stringly typed right now (you can pass all of the complex objects and numbers as strings programmatically)

sergeychernyshev (approved):
Looks great, thank you @43081j!

Just one minor comment to maybe add @thows as I was originally confused on return type for this function (e.g. why would it never fail?).

Can you rebase on the latest main and squash - and I'll merge it in.

sergeychernyshev:
When I squash merge, it looses automated attribution, but I can add a tag in the comment to attribute it to you still, I believe.

Let me try and we'll see how that goes ;)

sergeychernyshev:
I could not squash-merge without resolving the conflicts so I did it locally and merged through #50 - closing this one now.

Thanks a lot, @43081j, for helping out with this refactor!

sergeychernyshev:
Hmm... Looks like my conflict resolution went haywire... re-opening it back while I'm fixing that.

Issue #32, opened 2025-11-04
SSE OK


winkler1:
Some sites like https://checkboxes.andersmurphy.com/ keep an SSE channel open by design. This leads to a timeout (`page.goto: Timeout 5000ms exceeded.`) and only a .webm is captured. 

Can this be supported, perhaps by flag `--timeout-ok`? I've gotten it to work doing this. 
```  async doNavigation() {
    try {
      await this.page.goto(this.testURL, {
        waitUntil: 'networkidle'
      });
    } catch (error) {
      if (error.name === 'TimeoutError') {
        console.log('Navigation timeout occurred, continuing anyway...');
      } else {
        throw error;
      }
    }```

sergeychernyshev:
This should be added in 15e520e624763599314f7f11da2b2653ee7785aa as a default option, thank you, @winkler1 for bringing this up and @Giuseppe0804 for the PR.

It feel like making it a default behavior is a good idea as we'd rather get a report after timeout than not have one.

PR #37, opened 2025-11-06
fix: Close connection on timeout

Giuseppe0804:
https://github.com/cloudflare/telescope/issues/32

This could address websites that still keeps an open connection when `page.goto()` reaches its **timeout**(quite often if used with `{ waitUntil: 'networkidle' }`); also retroactively stops all connections by setting the browser context to offline.

### DIFF
--- a/lib/testRunner.js
+++ b/lib/testRunner.js
@@ -173,7 +173,16 @@ class TestRunner {
    * Triggers the navigation based on the passed in url, grabs a screenshot, and closes the context and browser
    */
   async doNavigation() {
-    await this.page.goto(this.testURL, { waitUntil: 'networkidle' });
+    try {
+      await this.page.goto(this.testURL, { waitUntil: 'networkidle' });
+    } catch (err) {
+      // If navigation timed out, set the context offline and continue.
+      if (err && (err.name === 'TimeoutError' || /Timeout/.test(err.message))) {
+        await this.page.context().setOffline(true);
+      } else {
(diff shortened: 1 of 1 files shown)


Giuseppe0804:
> Thank you for the PR, @Giuseppe0804 - do you feel there are some side-effects of doing this?

Thank you for raising the question about whether a test that fails to navigate to the page should be reported: https://github.com/cloudflare/telescope/issues/32

> In particular, if we treat it as not just a solution for SSE, but generic approach to generate reports for all long-running pages, do you foresee any issues with "going offline"?

This change is more of a workaround than a proper solution. I’m aware that it may affect pending operations(i.e: if the LCP element is an image that hasn’t finished loading and we go offline, it could create scenarios that don’t usually occur during normal navigation).

However, since Playwright doesn’t provide something similar to [Puppeteer’s lifecycleEvent like networkidle2](https://pptr.dev/api/puppeteer.puppeteerlifecycleevent)
, this approach could help trigger a more controlled version of the networkidle state (even though using networkidle is generally discouraged in the [Playwright documentation](https://playwright.dev/docs/api/class-page#page-goto)).

That said, I can imagine this wouldn’t be reliable in applications that keep persistent WebSocket connections or stream audio/video, as those would prevent the network from ever appearing “idle.”

> Do you know what happens to the in-flight requests?

They fail, i suppose that this would trigger `page.on('requestfailed')`

PR #64, opened 2025-12-09
Fully parse the block parameters.

ozcoder:
Parse the parameters to the `--block` command line option in a more full way.

### DIFF
--- a/__tests__/cli.test.js
+++ b/__tests__/cli.test.js
@@ -7,11 +7,12 @@ import { BrowserConfig } from '../lib/browsers.js';
 
 let testId;
 let outputLogs;
+let errLogs;
 let harJSON = null;
 let metrics = null;
 let config = null;
-let goodBrowser = null;
-let okBrowser = null;
+let singleBrowser = null; // A browser for tests that is known to be available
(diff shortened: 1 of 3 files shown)


__tests__/cli.test.js
```diff
@@ -117,3 +121,306 @@ describe.each(browsers)('Basic Test: %s', browser => {
...
+    }
+    config = retrieveConfig(testId, safeBrowser);
+  });
+
+  it('generates a Configuration file', async () => {
+    expect(config).toBeTruthy();
+  });
+
+  it('Do not block one and two', async () => {
+    expect(config.options.block).toEqual([ ]);
```
sergeychernyshev: This should probably throw an input error instead of just ignoring the parameter.

__tests__/cli.test.js
```diff
@@ -9,19 +9,22 @@ let testId;
 let outputLogs;
 let harJSON = null;
 let metrics = null;
+let config = null;
+let goodBrowser = null;
```
sergeychernyshev: Sorry, I don't understand what `goodBrowser` and `okBrowser` mean - what is this for?

PR #65, opened 2025-12-09
Change the everything route into blocked domain, blocks routes to not…

ozcoder:
… be so greedy.


sergeychernyshev (changes requested):

sergeychernyshev:
Let's also add tests to make sure they worked.

Issue #66, opened 2025-12-09
Tests fail when using `list` or `html` option


sergeychernyshev:
## Describe the bug

I get errors when running tests with `list` or `html` options:

```
{"error": "(intermediate value).resolve is not a function", "success": false}
```

This is likely

## Telescope info

Please include the following information about your environment:

**Telescope version**: 1.0.0
**Node version**: v22.21.1 and v20.10.0
**Operating system version**: MacOS

## Steps to reproduce

Add `html: true` to a programmatic test

```
      const result = await launchTest({
        url: 'https://www.example.com/',
        browser: 'chrome,
        debug: true,
        html: true,
      });
```

## Expected results

Playwright test should run successfully

## Observed results

Playwright test result object has error message:
```
{"error": "(intermediate value).resolve is not a function", "success": false}
```

## Additional context

It is likely related to finding `.ejs` templates used for both of those options.

PR #72, opened 2025-12-17
Fixed list and html approach and tests

mjkozicki:
Handles https://github.com/cloudflare/telescope/issues/66


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "start": "node cli.js",
-    "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --verbose --maxWorkers=1",
```
sergeychernyshev: `--maxWorkers=1` is needed here because if multiple browsers run at the same time, their temp environments conflict with each other.

package.json
```diff
@@ -12,7 +12,7 @@
...
+    "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --verbose",
```
sergeychernyshev:
```suggestion
    "test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" jest --verbose --maxWorkers=1",
```

`--maxWorkers=1` is needed here because if multiple browsers run at the same time, their temp environments conflict with each other sometimes.

sergeychernyshev (commented):
A few small changes.

__tests__/artifacts.test.js
```diff
@@ -127,7 +129,9 @@ describe.each(browsers)('Generated HTML artifacts (%s)', browser => {
       const indexPath = path.resolve(result.resultsPath, 'index.html');
       expect(fs.existsSync(indexPath)).toBe(true);
     } finally {
-      cleanup([path.resolve(result?.resultsPath)]);
+      if(!process.env.CI){
```
sergeychernyshev: Same as above

__tests__/artifacts.test.js
```diff
@@ -146,7 +150,9 @@ describe.each(browsers)('Generated list artifacts (%s)', browser => {
       indexPath = path.resolve(result.resultsPath, '..', 'index.html');
       expect(fs.existsSync(indexPath)).toBe(true);
     } finally {
-      cleanup([path.resolve(result?.resultsPath), indexPath]);
+      if(!process.env.CI){
```
sergeychernyshev: This we need in all environment, not just CI to make sure that when tested locally, each browser run regenerates the `results/index.html` every time.

PR #74, opened 2025-12-17
An npm script for running tests as CI does

sergeychernyshev:
To help testing CI test behavior locally.
- runs in headless mode
- test only in one browser right now (firefox)


joesanchezjr (approved):
@sergeychernyshev we should consider cross-env, but it's not a blocker for this

PR #77, opened 2025-12-19
move to packages folder

mjkozicki:
Reorg folder structure to follow standard for npm packages.


sergeychernyshev (changes requested):
Well, this would be too simple ;)

Let's make sure we keep the docs in-place, keep all the terminal commands working, move cli scripts and so on.

Also, we should probably keep the name `telescope` instead of `telescope-agent` as it will be the main NPM module.

PR #80, opened 2026-01-09
Provides device emulation via -d/--device flag using playwright

MildMax:
- Provides new flag -d/--device to allow device emulation
- Users are allowed to provide custom viewports that override device defaults via --width and --height CLI args

### DIFF
--- a/lib/chromeRunner.js
+++ b/lib/chromeRunner.js
@@ -1,9 +1,12 @@
 import { TestRunner } from './testRunner.js';
 import { log } from './helpers.js';
+import path from 'path';
+
 class ChromeRunner extends TestRunner {
   constructor(options, browserConfig) {
     //call parent
     super(options, browserConfig);
+    this.cdpClient = null;
   }
   /**
    * Given a browser instance, grab the page and then kick off anything that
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -222,16 +228,8 @@ export default function browserAgent() {
...
-      new Option('--width <int>', 'Viewport width, in pixels').default(
-        String(DEFAULT_OPTIONS.width),
-      ),
-    )
-    .addOption(
-      new Option('--height <int>', 'Viewport height, in pixels').default(
-        String(DEFAULT_OPTIONS.height),
-      ),
-    )
+    .addOption(new Option('--width <int>', 'Viewport width, in pixels'))
```
mjkozicki:
```suggestion
    .addOption(new Option('--width <int>', 'Viewport width, in pixels. If both width and device are provided, the width value will override device emulation viewport width.')).default(
        String(DEFAULT_OPTIONS.width),
      ),
```

index.js
```diff
@@ -292,7 +290,13 @@ export default function browserAgent() {
...
         '--dry',
         'Dry run (do not run test, just save config and cleanup)',
-      ).default(DEFAULT_OPTIONS.dry),
+      )
+    )
+    .addOption(
+      new Option(
+        '-d, --device <string>',
+        'Device to emulate; devices are based on the Playwright device list (see https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json)',
+      )
```
mjkozicki: Add default option for device.

lib/config.js
```diff
@@ -29,6 +28,42 @@ export function normalizeCLIConfig(options) {
...
+    config.width = parseInt(options.width);
+  } else if (
+    config.device &&
+    config.device.viewport &&
+    config.device.viewport.width
+  ) {
+    config.width = config.device.viewport.width;
+  } else {
+    config.width = DEFAULT_OPTIONS.width;
+  }
```
mjkozicki: Does the `config.width` need to be set if there is no override for the `config.device.viewport.width` value?

```suggestion
  // set width and height from options before assigning device viewport or defaults
  if (options.width) {
    config.width = parseInt(options.width);
  }
```

lib/defaultOptions.js
```diff
@@ -37,4 +37,6 @@ export const DEFAULT_OPTIONS = {
   zip: false,
   // Dry run (false = no dry run)
   dry: false,
+  // Device to emulate (empty string = no emulation)
+  device: '',
```
mjkozicki: It would be clearer is device defaulted to `null`.

Issue #58, opened 2025-11-20
Add option for upload


mjkozicki:
Depends on #42 

Add an option to upload the results via API to an arbitrary endpoint. Currently assume only the zip file will be uploaded and whatever usage on the backend to unip and store will be an independent decision.

`--upload <url>. Provide a URL for the results to be POSTed to`

Options if `<url>` is provided,
1. zip the files regardless of whether `--zip` is specified, upload, then delete zip
2. zip the files regardless of whether `--zip` is specified, upload, then keep zip

If `--zip` and `--url` are specified, always zip and upload.

PR #84, opened 2026-01-20
Add uploadUrl option

mjkozicki:
Closes https://github.com/cloudflare/telescope/issues/58

### DIFF
--- a/__tests__/artifacts.test.js
+++ b/__tests__/artifacts.test.js
@@ -131,9 +131,7 @@ describe.each(browsers)('Generated HTML artifacts (%s)', browser => {
       const indexPath = path.resolve(result.resultsPath, 'index.html');
       expect(fs.existsSync(indexPath)).toBe(true);
     } finally {
-      if (!process.env.CI) {
-        cleanup([path.resolve(result.resultsPath)]);
-      }
+      cleanup([path.resolve(result.resultsPath)]);
     }
   }, 120000);
 });
@@ -153,9 +151,7 @@ describe.each(browsers)('Generated list artifacts (%s)', browser => {
       indexPath = path.resolve(result.resultsPath, '..', 'index.html');
       expect(fs.existsSync(indexPath)).toBe(true);
     } finally {
-      if (!process.env.CI) {
-        cleanup([path.resolve(result.resultsPath), indexPath]);
-      }
+      cleanup([path.resolve(result.resultsPath), indexPath]);
     }
   }, 120000);
 });
@@ -187,11 +183,9 @@ describe.each(browsers)('Upload zip for browsers (%s)', browser => {
         expect(result.success).toBe(true);
         expect(fs.existsSync(zipfile)).toBe(zip);
       } finally {
-        if (!process.env.CI) {
-          cleanup([path.resolve(result.resultsPath)]);
-          if (zip) {
-            cleanup([zipfile]);
-          }
+        cleanup([path.resolve(result.resultsPath)]);
+        if (zip) {
+          cleanup([zipfile]);
         }
       }
     });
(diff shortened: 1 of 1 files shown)


lib/browsers.js
```diff
@@ -168,6 +168,18 @@ class BrowserConfig {
       browserConfig.viewport.height = options.height;
       browserConfig.recordVideo.size.height = options.height;
     }
+
+    // Validate uploadUrl if provided
+    if (options.uploadUrl) {
+      try {
+        new URL(options.uploadUrl);
+      } catch (err) {
+        const err_obj = Error(`--uploadUrl must be a valid URL`);
```
sergeychernyshev:
```suggestion
        throw new Error(`--uploadUrl must be a valid URL`);
```

__tests__/artifacts.test.js
```diff
@@ -156,3 +159,77 @@ describe.each(browsers)('Generated list artifacts (%s)', browser => {
...
+      }
+    }
+  });
+});
+
+describe('Invalid upload URL', () => {
+  test('Error when invalid --uploadUrl is specified.', async () => {
+    const config = {
+      url: 'https://www.example.com/',
+      browser: 'firefox',
```
sergeychernyshev: Instead of hardcoding a specific browser to pass CI tests, you should use `describe.each(browsers)(....)` setup like other test above.

It uses the `const browsers = BrowserConfig.getBrowsers();` call to retrieve the browsers and in CI environment, it will only return `['firefox']` (e.g. we only support Firefox in CI right now).

You can emulate that locally on your dev machine by setting `CI` environment variable to `true` before running the test like so:

```
CI=true npm test artifacts
```
(it also forces playwright to go into the headless mode)

lib/testRunner.js
```diff
@@ -602,9 +604,41 @@ class TestRunner {
       }
     }
 
+    // handle uploading
+    if (this.options.uploadUrl) {
+      try {
+        // if file hasn't been zipped yet, do it now
+        if (!existsSync(outputZip)) {
```
sergeychernyshev: This block is redundant with `--zip` option - let's reuse the code, e.g. force `--zip` when `--uploadUrl` is set.

__tests__/artifacts.test.js
```diff
@@ -156,3 +159,77 @@ describe.each(browsers)('Generated list artifacts (%s)', browser => {
...
+        zip: zip,
+      };
+      result = await launchTest(config);
+      zipfile = path.resolve(resultsRoot, `${result.testId}.zip`);
+
+      expect(result).toBeDefined();
+      expect(result.success).toBe(true);
+      expect(fs.existsSync(zipfile)).toBe(zip);
+    } finally {
+      if (!process.env.CI) {
```
sergeychernyshev: Why not cleanup in CI? any particular reason?

__tests__/artifacts.test.js
```diff
@@ -156,3 +159,77 @@ describe.each(browsers)('Generated list artifacts (%s)', browser => {
...
+  afterEach(() => server.resetHandlers()); // Reset any runtime handlers (prevents test cross-contamination)
+  afterAll(() => server.close()); // Clean up once all tests are done
+  test('POST when --uploadUrl is specified.', async () => {
+    let result, zipfile;
+    try {
+      let config = {
+        url: 'https://www.example.com/',
+        browser: 'firefox',
+        uploadUrl: 'https://api.example.com/upload',
+        zip: zip,
```
sergeychernyshev:
```suggestion
        zip,
```

sergeychernyshev (changes requested):
Hacking our own CI setup is not a solution, Mike ;)

Few more requests here and there.

mjkozicki:
Reworked to `upload` branch. Superceded by https://github.com/cloudflare/telescope/pull/146

PR #102, opened 2026-01-29
Initial Upload Page

Judyzc:
This is the first version of the upload page for telescopetest-io. This would add the page telescopetest.io/upload. 
Additions:
- upload.ts: mainly a POST: 
    - D1 connection with schema outlined in db/schema.sql, already generated using wrangler commands. This holds metadata for the test cards on the results overview page for later. Each entry has its own generated test_id, as well as a unique hash for the contents of the zip folder to prevent duplicates.
    - R2 connection. This holds all the unzipped files from a zip upload, prefixed by '{test_id}/'. 
- Upload.astro: calls on upload.ts. Uses TopNav.astro and Layout.astro. Has night mode and link to github. 
- types/testConfig.ts: interface for testConfig object for D1.
- Files in src/lib: classes for using D1 and R2. 

Still needed/TODO (*Edit: these are additional issues for after this PR): 
- Use prisma ORM 
- Build and test uploading just the POST endpoint (not through UI)  
- Start looking at manifest for handing R2 unzipped info
- Creating staging env different from prod
- Make light/dark mode toggle not based on javascript 
- Write up initialization setup in README
- Clean up some engine/pngs

### DIFF
--- a/telescopetest-io/src/layouts/Layout.astro
+++ b/telescopetest-io/src/layouts/Layout.astro
@@ -25,7 +25,7 @@ const {
   </body>
 </html>
 
-<style>
+<style is:global>
   :root {
     color-scheme: light;
     font-family:


telescopetest-io/src/pages/api/upload.ts
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Extract file list from ZIP archive
+ * Works in both Node.js (adm-zip) and Cloudflare Workers (fflate) environments
+ * @param buffer - ArrayBuffer containing ZIP file data
+ * @returns Promise<string[]> - Array of file paths/names in the ZIP
+ */
+async function getFilesFromZip(
+  buffer: ArrayBuffer,
+): Promise<Record<string, any>> {
+  const { unzipSync } = await import('fflate');
```
sergeychernyshev: This package is not included in package.json.
This causes `npm run build` to fail.

telescopetest-io/src/lib/d1/repositories/test-repository.ts
```diff
@@ -0,0 +1,57 @@
+import { D1Client } from '../d1-client';
+import { TestConfig } from '../../../types/testConfig';
+export class TestRepository {
```
sergeychernyshev: Let's not use the term "Repository" for this - it might be confusing as the word most often refers to version control repository, e.g. git or subversion.

`TestStore` might be a better name.

We'll also make `TestStore` abstract in the future and implement `LocalTestStore` so maybe call this one `D1TestStore`.

telescopetest-io/src/lib/d1/repositories/test-repository.ts
```diff
@@ -0,0 +1,57 @@
...
+  private client: D1Client;
+  constructor(d1Binding: any) {
+    this.client = new D1Client(d1Binding);
+  }
+  /**
+   * Create a new test entry
+   * Will throw error if duplicate zip_key
+   */
+  async create(testConfig: TestConfig) {
```
sergeychernyshev:
```suggestion
  async createTest(testConfig: TestConfig) {
```

Let's make the method with the rest of the methods like `findTest...` etc.

Otherwise it might be confused with a static factory for TestRepository itself which is not necessary as we use `constructor()` for that.

telescopetest-io/src/pages/api/upload.ts
```diff
@@ -0,0 +1,146 @@
...
+  const hashBuffer = await crypto.subtle.digest('SHA-256', buffer);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  return hashHex;
+}
+
+
+export const POST: APIRoute = async (context: any) => {
+  try {
+    const formData = await context.request?.formData();
```
sergeychernyshev: I assume `request` property is always there, let's make sure `context` is typed (see above) and then you don't need a question mark here.

telescopetest-io/src/pages/api/upload.ts
```diff
@@ -0,0 +1,146 @@
...
+      return new Response(
+        JSON.stringify({
+          success: false,
+          error: 'No config.json file found in the ZIP archive',
+        }),
+        { status: 402, headers: { 'Content-Type': 'application/json' } },
+      );
+    }
+    // Extract and parse config.json
+    const configData = unzipped[configFile];
```
sergeychernyshev: I'd rename `configData` to `configBytes` as it can be confused with parsed data, e.g. `config` below.

So overall flow:
`configBytes` -(utf-8 decoding)-> `configText` -(JSON parse)-> `config`

telescopetest-io/src/pages/api/upload.ts
```diff
@@ -0,0 +1,146 @@
...
+      return new Response(
+        JSON.stringify({
+          success: false,
+          error: 'Failed to extract config.json from ZIP',
+        }),
+        { status: 403, headers: { 'Content-Type': 'application/json' } },
+      );
+    }
+    const configText = new TextDecoder('utf-8').decode(configData);
+    const config = JSON.parse(configText);
```
sergeychernyshev: `JSON.parse()` call might fail too if config is invalid.

Let's catch it as well.

telescopetest-io/src/pages/api/upload.ts
sergeychernyshev: Also, let's use `zod` library formData validation.

sergeychernyshev (changes requested):
Overall looks good, but have a few small items here and there.

telescopetest-io/README.md
```diff
@@ -1,3 +1,37 @@
...
+This is the website for users to upload and view telescope ZIP results. This is built with Astro web framework and hosted on Cloudflare Workers.
+
+## Work Log
+
+- Rebuilt a default astro workers project and moved all my edits into there. Ran with `npm create cloudflare@latest -- telescopetest-io --framework=astro`: https://developers.cloudflare.com/workers/framework-guides/web-apps/astro/#deploy-a-new-astro-project-on-workers
+- Added custom domain (astro.telescopetest.io) on Cloudflare Workers UI.
+- IMPORTANT: had to disable null rule on telescopetest.io domain to have css render properly
+
+- To create a d1 db, I ran `npx wrangler@latest d1 create telescope-db`: https://developers.cloudflare.com/d1/get-started/.
+- Then connected it to a db/schema.sql I made: `npx wrangler d1 execute telescope-db --local --file=./db/schema.sql`. Got some error about FileHandle error 22. This created a local db. I had to fix the error by updating my local wrangler version to the most recent.
```
sergeychernyshev: I ended up needing to run this command in order for the app to run locally, but the other command said that remote database already exists.

telescopetest-io/src/components/TopNav.astro
```diff
@@ -0,0 +1,58 @@
...
+export interface Props {
+  active?: string;
+}
+
+const { active } = Astro.props;
+
+const navItems = [
+  { name: 'home', label: 'Home', href: '/' },
+  { name: 'upload', label: 'Upload', href: '/upload' },
+];
```
joesanchezjr: Instead of having this "active" prop, we can leverage the Astro render context and use `Astro.url.pathname`. This would simplify the navItems array so the objects don't need the name property. Then below you can compare against `item.href`.

Reference: https://docs.astro.build/en/reference/api-reference/#url

telescopetest-io/src/lib/d1/d1-client.ts
```diff
@@ -0,0 +1,27 @@
+export class D1Client {
+  private db: any; // D1Database binding
+  constructor(db: any) {
+    this.db = db;
+  }
```
joesanchezjr: Similar to my note above, this type should already be gnerated

telescopetest-io/src/pages/api/upload.ts
```diff
@@ -0,0 +1,187 @@
...
+ * Generate a SHA-256 hash of the buffer contents to use as unique identifier
+ * @param buffer - ArrayBuffer containing the file data
+ * @returns Promise<string> - Hex string of the hash
+ */
+async function generateContentHash(buffer: ArrayBuffer): Promise<string> {
+  const hashBuffer = await crypto.subtle.digest('SHA-256', buffer);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map(b => b.toString(16).padStart(2, '0')).join('');
+  return hashHex;
+}
```
joesanchezjr: @sergeychernyshev do you know why we need to use crypto here and not like uuid or some other unique id?

telescopetest-io/src/pages/api/upload.ts
```diff
@@ -49,16 +70,15 @@ export const POST: APIRoute = async (context: any) => {
     const files = Object.keys(unzipped).filter(name => !name.endsWith('/'));
```
sergeychernyshev: I didn't catch it first time, but it looks like this filter is either unnecessary, or better be moved to `getFilesFromZip()` function. More over, it looks like `unzipSync()` actually supports filter function as a second argument so we don't get files we don't want to begin with.

telescopetest-io/src/pages/upload.astro
```diff
@@ -0,0 +1,365 @@
+---
+import Layout from '../layouts/Layout.astro';
+import TopNav from '../components/TopNav.astro';
+
+export const prerender = false;
```
joesanchezjr: this page isn't using any dynamic server side stuff that would warrant it not be prerendered, so we should remove this line

joesanchezjr (changes requested):

telescopetest-io/src/pages/api/upload.ts
```diff
@@ -74,53 +94,74 @@ export const POST: APIRoute = async (context: any) => {
...
         break;
     }
     // store the test config (metadata) in the db
-    await testRepo.create(testConfig);
-    const testId = await testRepo.findTestIdByZipKey(zipKey);
+    await testStore.createTestFromConfig(testConfig);
+    const testId = testConfig.test_id; // generated in constructor
+    console.log('test id: ', testId);
     if (!testId) {
       throw new Error('Failed to retrieve test_id after insert');
```
sergeychernyshev: This sounds like you expected `createTestFromConfig` to update the `testConfig` object, but in reality that was done earlier in constructor so this error doesn't really tell you if insert was successful or not.

I believe here you should just remove `testId` check altogether (we know it comes from constructor right now) and check if whatever `await testStore.createTestFromConfig(testConfig);` returned a success or not (don't know if it is supposed to throw an error on a failed SQL INSERT).

When it is time to implement auto-increment (e.g. DB generating IDs), you can change `createTestFromConfig` to either return the ID generated based on SQLite `last_row_id`, or to return a new `TestConfig` object that has an ID.

telescopetest-io/src/layouts/Page.astro
```diff
@@ -0,0 +1,17 @@
...
+  <TopNav />
+  <slot />
+</Layout>
+
+<style>
+  :global(.data-content) {
+    max-width: 1400px;
+    margin: 0 auto;
+    padding: 24px;
+  }
```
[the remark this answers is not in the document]
Judyzc: Yeah I ended up just changing this. I saw your comment just now sorry!

telescopetest-io/package.json
```diff
@@ -6,16 +6,20 @@
...
+    "astro": "^5.16.15",
+    "fflate": "^0.8.2",
+    "zod": "^3.25.76"
   },
   "devDependencies": {
     "@types/node": "^25.0.10",
     "wrangler": "^4.61.0"
+  },
+  "engines": {
+    "node": ">18.17.1"
```
sergeychernyshev: There is no need for this - `astro` package itself would declare it's engines dependency - there is no dependency in our own code.

telescopetest-io/src/pages/upload.astro
```diff
@@ -2,189 +2,72 @@
...
+        <div class="file-upload-area" id="file-upload-area">
+          <input
+            type="file"
+            id="file-upload"
+            name="file"
+            accept=".zip"
+            style="display: none;"
+          />
+          <img
+            src="/upload.png"
```
sergeychernyshev: There is no need to replace SVG with PNG for regular icons - the not supporting SVG limitation only applies to favorite icons.

We can keep the upload icon displayed in the body of the page as SVG.

Issue #104, opened 2026-02-04
Typechecking


sufian-cf:
Telescope has JSDoc type annotations which have diverged from reality.

It'd be good to have TypeScript perform type checking on the JavaScript code using these annotations, checks running in github, and type errors down to zero.

PR #113, opened 2026-02-06
Typescript refactor

thatzopoulos:
All .js files converted to .ts:
   - lib/\*.js → src/\*.ts
   - index.js → src/index.ts
   - cli.js → src/cli.ts
   - \_\_tests\_\_\/*.test.js → \_\_tests\_\_\/*.test.ts
   - \_\_tests\_\_\/helpers.js → \_\_tests\_\_\/helpers.ts 

New TypeScript Config Files:
   - tsconfig.json, jest.config.js, eslint.config.js, src/types.ts, src/ffmpeg.d.ts, src/sitespeed-throttle.d.ts

package.json changes: 
- Added build scripts, changed entry points to dist/, added TypeScript devDependencies

Other:
- All CLI tests now reference dist/src/cli.js instead of cli.js
- Updated CLI Args to use slice (non-mutating) instead of splice (mutating)
- Cleanup test method in src/testRunner.ts is now async and closes the browser instance before removing temp files to prevent Jest open handles

- Resolves https://github.com/cloudflare/telescope/issues/104

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -66,7 +66,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.4.tgz",
       "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
       "dev": true,
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
@@ -1291,7 +1290,6 @@
       "integrity": "sha512-CPrnr8voK8vC6eEtyRzvMpgp3VyVRhgclonE7qYi6P9sXwYb59ucfrnmFBTaP0yUi8Gk4yZg/LlTJULGxvTNsg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~7.16.0"
       }
@@ -1352,7 +1350,6 @@
       "integrity": "sha512-BtE0k6cjwjLZoZixN0t5AKP0kSzlGu7FctRXYuPAm//aaiZhmfq1JwdYpYr1brzEspYyFeF+8XF5j2VK6oalrA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.54.0",
         "@typescript-eslint/types": "8.54.0",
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -37,6 +37,9 @@ jobs:
         # don't install the browsers since we're using the pre-built container
         run: npm ci --ignore-scripts
 
+      - name: Build TypeScript
```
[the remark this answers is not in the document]
thatzopoulos-cf: It worked

src/config.ts
```diff
@@ -1,30 +1,35 @@
...
     debug: options.debug || DEFAULT_OPTIONS.debug,
     html: options.html || DEFAULT_OPTIONS.html,
     openHtml: options.openHtml || DEFAULT_OPTIONS.openHtml,
     list: options.list || DEFAULT_OPTIONS.list,
-    connectionType: options.connectionType || DEFAULT_OPTIONS.connectionType,
-    auth: options.auth || DEFAULT_OPTIONS.auth,
+    connectionType:
+      (options.connectionType as ConnectionType) ||
+      DEFAULT_OPTIONS.connectionType,
+    auth: DEFAULT_OPTIONS.auth,
```
sergeychernyshev: Looks like you dropped `options.auth` here.

src/config.ts
```diff
@@ -39,7 +44,7 @@ export function normalizeCLIConfig(options) {
   }
 
   if (options.auth) {
-    config.auth = JSON.parse(options.auth);
+    config.auth = JSON.parse(options.auth) as HTTPCredentials;
```
sergeychernyshev: I wonder if we should start using `zod` for input validation and type definitions here.

Not really a blocker for this initial PR, but might be useful as just casting to type is very optimistic ;)

src/config.ts
```diff
@@ -1,30 +1,35 @@
...
+export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
+  const config: LaunchOptions = {
     url: options.url,
-    browser: options.browser || DEFAULT_OPTIONS.browser,
-    width: parseInt(options.width) || DEFAULT_OPTIONS.width,
-    height: parseInt(options.height) || DEFAULT_OPTIONS.height,
-    frameRate: parseInt(options.frameRate) || DEFAULT_OPTIONS.frameRate,
-    timeout: parseInt(options.timeout) || DEFAULT_OPTIONS.timeout,
+    browser:
+      (options.browser as LaunchOptions['browser']) || DEFAULT_OPTIONS.browser,
```
sergeychernyshev:
```suggestion
      (options.browser as BrowserName || DEFAULT_OPTIONS.browser,
```

src/config.ts
```diff
@@ -1,30 +1,35 @@
...
+    width: parseInt(String(options.width)) || DEFAULT_OPTIONS.width,
```
sergeychernyshev: Hmm. We're casting `options.width` to `string` which looses optionality (e.g. `undefined`).

Is that a problem here? Does wrapping into String constructor even work? and should that be `new String()` instead?

We might need to write some tests for this to make sure it still works.

sergeychernyshev (approved):
None of these are critical changes, I believe so if you feel like making follow-up tickets for improving types so we can deal with it later, it's fine with me.

src/config.ts
thatzopoulos-cf: String(undefined) returns "undefined" and parseInt("undefined") returns NaN but this could definitely be refactored.

PR #127, opened 2026-02-12
telescopetest.io: Add Prisma ORM for D1

Judyzc:
- Added new prisma packages and set up prisma client. 
- Moved the initial db/schema.sql into a proper migrations/0001_init.sql that is now tracked by Wrangler as a migration. 
- Deleted previous D1 helper functions and now moved to using Prisma for duplicate check and inserting new record 
- Documented new prisma steps and future migration steps
- Edited DEPLOY.yml workflow to add in migrations and prisma client generation before build 
- Tested on staging


telescopetest-io/src/pages/api/upload.ts
```diff
@@ -149,16 +175,17 @@ export const POST: APIRoute = async (context: APIContext) => {
...
     }
-    const testId = testConfig.test_id; // generated in constructor
     // store all unzipped files in R2 with {testId}/{filename} format
-    // storing with {testId}/ for future expansion to multiple users
     for (const filename of files) {
       await env.RESULTS_BUCKET.put(`${testId}/${filename}`, unzipped[filename]);
     }
+    // disconnect from prisma
+    await prisma.$disconnect();
+    // return success
```
joesanchezjr: Since we're running in a worker instead of a long-lived service like Node, we don't need to use `$disconnect()` - two spots in this file. https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration/databases-connections#prismaclient-in-serverless-environments

PR #130, opened 2026-02-16
Command line options to change User-Agent string

### LINKED ISSUE #17: Feature: Append to UA
It would be nice to append to the UA string, particularly useful for authentication purposes.

ozcoder:
Copy the agent options to browserConfig.
They default to undefined.
Added command line options --userAgent and --agentExtra 
Open a blank page to get the original User-Agent string. 
Added them to the needed type definitions.

Closes #17


src/browsers.ts
```diff
@@ -208,6 +208,13 @@ class BrowserConfig {
       browserConfig.recordVideo.size.height = options.height;
     }
 
+    if (options.userAgent) {
+      browserConfig.userAgent = options.userAgent;
+    }
+    if (options.agentExtra) {
```
sergeychernyshev: I don't think we need to set `agentExtra` on the object that we pass to playwright - we just use it in testAgent to modify the `userAgent` string later.

src/testRunner.ts
```diff
@@ -170,6 +170,22 @@ class TestRunner {
     const engine = this.selectedBrowser.engine;
     const browserType = playwright[engine];
 
+    if (this.options.agentExtra) {
+      const tmpbrowser = await browserType.launch(
+        this.selectedBrowser as Parameters<
```
sergeychernyshev: Basically instead of passing the whole `this.selectedBrowser` pass a version of it with minimal number of options set.

Or clone `this.selectedBrowser` and unset the heavy options so we can get the most accurate userAgent string, but remove as much overhead as possible.

src/types.ts
```diff
@@ -128,7 +128,7 @@ export interface BrowserConfigOptions {
   channel?: BrowserChannel;
   headless: boolean;
   viewport: { width: number; height: number };
-  recordHar: { path: string };
+  recordHar?: { path: string };
```
sergeychernyshev: Let's keep it a required property on `BrowserConfigOptions` type which we'll use for full testing.

And create a separate type for the simplified version, e.g. `SimplifiedBrowserConfigOptions` and remove the props we don't want in it.

You can use Omit to just remove the props we want to unset:
```
type NewType = Omit<OriginalType, 'prop1' | 'prop2'>;
```

src/types.ts
```diff
@@ -128,7 +128,7 @@ export interface BrowserConfigOptions {
...
   recordVideo: { dir: string; size: { width: number; height: number } };
```
sergeychernyshev: We probably want this excluded in simplified type as well.

src/testRunner.ts
```diff
@@ -74,9 +74,14 @@ class TestRunner {
...
 
-    this.selectedBrowser.recordHar.path =
-      this.paths['results'] + '/pageload.har';
-    this.selectedBrowser.recordVideo.dir = this.paths['results'];
+    if (this.selectedBrowser.recordHar) {
+      this.selectedBrowser.recordHar.path =
+        this.paths['results'] + '/pageload.har';
+    }
+
+    if (this.selectedBrowser.recordVideo) {
```
sergeychernyshev: Same as above - no need for the wrapper.

sergeychernyshev (changes requested):

src/types.ts
ozcoder: OK

src/testRunner.ts
```diff
@@ -171,8 +176,23 @@ class TestRunner {
     const browserType = playwright[engine];
 
     if (this.options.agentExtra) {
+      let simpleOptions = JSON.parse(JSON.stringify(this.selectedBrowser)); // Deep clone
```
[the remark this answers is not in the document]
ozcoder: I tried a bunch of techniques and couldn't get the copy to work the way I wanted.
I'm just manually creating the simplified options. At least it is very obvious what is happening now.

src/testRunner.ts
```diff
@@ -74,9 +74,14 @@ class TestRunner {
     this.paths['filmstrip'] = this.paths.results + '/filmstrip';
     mkdirSync(this.paths['results'], { recursive: true });
 
-    this.selectedBrowser.recordHar.path =
-      this.paths['results'] + '/pageload.har';
-    this.selectedBrowser.recordVideo.dir = this.paths['results'];
+    if (this.selectedBrowser.recordHar) {
```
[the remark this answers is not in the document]
ozcoder: OK

__tests__/cli.test.ts
```diff
@@ -68,3 +74,117 @@ describe.each(browsers)('Basic Test: %s', browser => {
...
+
+  beforeAll(() => {
+    try {
+      const args = [
+        'node',
+        'dist/src/cli.js',
+        '--url',
+        'https://www.example.com',
+        '--agentExtra',
+        ' TELESCOPE_TEST',
```
sergeychernyshev:
```suggestion
        agentExtra,
```

src/types.ts
```diff
@@ -125,6 +125,7 @@ export type BrowserConfigEntry =
  * Full browser configuration with all runtime options
  */
 export interface BrowserConfigOptions {
+  agentExtra?: string;
```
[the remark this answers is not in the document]
ozcoder: Yes, you are right. Removed.

src/testRunner.ts
```diff
@@ -211,6 +212,39 @@ class TestRunner {
...
+        channel: this.selectedBrowser.channel,
+        engine: this.selectedBrowser.engine,
+        firefoxUserPrefs: this.selectedBrowser.firefoxUserPrefs,
+        headless: this.selectedBrowser.headless,
+        ignoreDefaultArgs: this.selectedBrowser.ignoreDefaultArgs,
+        viewport: { height: 1, width: 1 },
+      };
+
+      if (simpleOptions.args) {
+        let idx = simpleOptions.args.indexOf('--metrics-recording-only');
```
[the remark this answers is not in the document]
ozcoder: OK

__tests__/cli.test.ts
```diff
@@ -1,9 +1,12 @@
...
 
-import { retrieveHAR, retrieveMetrics } from './helpers.js';
+import { retrieveHAR, retrieveMetrics, cleanupTestDirectory } from './helpers.js';
 
 import { BrowserConfig } from '../src/browsers.js';
-import type { HarData, Metrics } from '../src/types.js';
+import type { HarData, Metrics, HTTPHeader } from '../src/types.js';
+
+import { expect } from '@jest/globals';
+import './endsWith.ts'; // Extending expect with endsWith method
```
sergeychernyshev: No need for a new matcher
```suggestion
```

__tests__/endsWith.ts
sergeychernyshev: I don't think we need a new matcher ;)

__tests__/cli.test.ts
```diff
@@ -68,3 +75,119 @@ describe.each(browsers)('Basic Test: %s', browser => {
...
+    if (harJSON) {
+      const htmlUserAgent = harJSON
+        .log
+        .entries[0]
+        .request
+        .headers
+        .find((hdr: HTTPHeader) => hdr.name === 'User-Agent');
+
+      if (htmlUserAgent) {
+        expect(htmlUserAgent.value).endsWith(agentExtraText);
```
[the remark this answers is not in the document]
ozcoder: Wow. I didn't know about `endsWith` and `startsWith` on `String`. That's what decades of supporting Internet Explorer do to my brain.
Much easier than writing a new matcher.

PR #131, opened 2026-02-17
Fade in the sections with CSS animation.

ozcoder:
Do not add the visible class if using the CSS animation.
Handle reduced motion users with a CSS check

### DIFF
--- a/telescopetest-io/src/pages/index.astro
+++ b/telescopetest-io/src/pages/index.astro
@@ -400,46 +400,33 @@ const year = new Date().getFullYear();
         border: 1px solid rgba(139, 58, 140, 0.3);
         box-shadow: 0 25px 70px rgba(0, 0, 0, 0.4);
         position: relative;
-        opacity: 0;
-        transform: translateY(50px);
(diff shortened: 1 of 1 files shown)


ozcoder:
Sorry @sufian-cf , I'm still working on this. I was hoping someone would have a quick look and make sure I'm on the correct track.

sufian-cf:
No problem, just wanted to check the status. Next time if you are waiting for feedback, could you please let us know?

Was your specific goal to eliminate the IntersectionObserver, or change the animation to have the fadein linked to scroll? If it's the latter, this approach will not work in Firefox support, as the `animation-timeline` property is not supported at this time: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation-timeline

ozcoder:
I have removed the JavaScript fallback code and worked on the CSS animation a little bit.
I developed on Chrome 145 and tested with Edge 140. On Firefox 148 there is no animation, the _zones_ are visible and there are no errors in the console.
Sorry, I haven't checked with Safari yet.

PR #132, opened 2026-02-17
Provides device emulation via --deviceName flag using playwright

### LINKED ISSUE #4: Feature: 📱Mobile Emulation
We definitely need to get some mobile emulation in place. Playwright has a long list of device descriptors (https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json) which we could expose, or we could go with our own subset.

Either way, will want to make sure there's a good default `cpuThrottle` level applied for each as well.

MildMax:
- Provides new flag --device to allow device emulation
- Users are allowed to provide custom viewports that override device defaults via --width and --height CLI args
- Updates playwright to 1.58.2 as part of bug fix for isMobile spawning multiple ffmpeg processes on chromium browsers in 1.57.0
- Omits chrome-canary and chrome-beta from device emulation unit tests, adds note to README indicating telescope does not support device emulation for the chrome beta and chrome canary browsers
- Sets node version 24 to help with mobile emulation in chrome browsers

Closes #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Options:
   -c, --cookies <object>        Any custom cookies to apply
   -f, --flags <string>          A comma separated list of Chromium flags to launch Chrome with. See: https://peter.sh/experiments/chromium-command-line-switches/
   --agentExtra <string>         A string to append to the end of the default User-Agent. Takes precedence over --userAgent option.
+  --device '<string>'           Device to use device emulation (viewport size, DPR, touch events); devices are based on the Playwright device list (see https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json)
   --blockDomains <domains...>   A comma separated list of domains to block
   --block <substrings...>       A comma-delimited list of urls to block (based on a substring match)
   --delay <object>              An object mapping request regexes to response delays. Example: '{".css$": 2000, ".js$": 5000}'
--- a/__tests__/emulation.test.ts
+++ b/__tests__/emulation.test.ts
@@ -3,6 +3,7 @@ import type { SuccessfulTestResult } from '../src/index.js';
 import { launchTest } from '../src/index.js';
 import { BrowserConfig } from '../src/browsers.js';
 import { normalizeCLIConfig } from '../src/config.js';
+import type { LaunchOptions, BrowserConfigOptions } from '../src/types.js';
 import { describe, it, expect, beforeAll } from 'vitest';
 
 const browsers = BrowserConfig.getBrowsers();
@@ -22,23 +23,28 @@ describe.each(browsers)(
     describe.each(target_devices)(
       'Setting device emulation updates the config for device: %s',
       device => {
-        let options = {
-          browser,
-          device: device,
-          url: '../tests/sandbox/index.html',
-        };
-        var config_options = normalizeCLIConfig(options);
+        let config_options: LaunchOptions;
+        let config: BrowserConfigOptions;
+
+        beforeAll(() => {
+          config_options = normalizeCLIConfig({
+            browser,
+            device: device,
+            url: '../tests/sandbox/index.html',
+          });
+          config = new BrowserConfig().getBrowserConfig(
+            browser,
+            config_options,
+          );
+        });
+
         it(`Test device emulation does not set default width and height values for browser: ${browser}`, () => {
           expect(config_options && typeof config_options === 'object').toBe(
             true,
           );
-          expect(config_options.width).toBe(false);
-          expect(config_options.height).toBe(false);
+          expect(config_options.width).toBeUndefined();
+          expect(config_options.height).toBeUndefined();
         });
-        let config = new BrowserConfig().getBrowserConfig(
-          browser,
-          config_options,
-        );
         it(`Setting device emulation creates a valid config object for browser: ${browser}`, () => {
           expect(config && typeof config === 'object').toBe(true);
         });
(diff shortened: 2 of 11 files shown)


sergeychernyshev:
I also see extra space, please run the `npm run format` on it.

sergeychernyshev (commented):
All looks great and straightforward with just a couple comments.

Still getting
```
Jest did not exit one second after the test run has completed.

'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Judyzc:
> All looks great and straightforward with just a couple comments.
> 
> Still getting
> 
> ```
> Jest did not exit one second after the test run has completed.
> 
> 'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
> ```

I got the same thing with jest and that's about it

Judyzc:
> > All looks great and straightforward with just a couple comments.
> > Still getting
> > ```
> > Jest did not exit one second after the test run has completed.
> > 
> > 'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
> > ```
> 
> I got the same thing with jest and that's about it

I'm no longer getting this timeout warning

ryantownsend (commented):
Couple of very minor documentation suggestions.

ryantownsend (approved):

__tests__/emulation.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+//test for other options
+describe.each(browsers)(
+  'Device Emulation CLI + Config tests using device properties',
+  browser => {
+    if (browser === 'canary' || browser === 'chrome-beta') {
+      console.info(`Skipping device emulation tests for browser: ${browser}`);
+      it.skip('Skipped for canary/beta', () => {});
+      return;
+    }
```
sufian-cf: Mostly curious: why are we skipping these?

src/index.ts
```diff
@@ -265,14 +265,16 @@ export default function browserAgent(): void {
...
+      ).default(DEFAULT_OPTIONS.width),
     )
     .addOption(
-      new Option('--height <int>', 'Viewport height, in pixels')
-        .default(DEFAULT_OPTIONS.height)
-        .argParser((v) => parseNumeric(PositiveIntSchema, v, '--height')),
+      new Option(
+        '--height <int>',
+        'Viewport height, in pixels. If both height and device are provided, the height value will override device emulation viewport height.',
+      ).default(DEFAULT_OPTIONS.height),
```
sufian-cf: Was dropping the validation here (.argParser, ensuring options were positive integers) intentional?

__tests__/emulation.test.ts
MildMax: ah sorry -- this is a leftover from when we had thought the issue was with beta and canary chrome versions, and so omitted them from the tests, I'll remove these skips

__tests__/artifacts.test.ts
```diff
@@ -215,7 +215,7 @@ describe.each(browsers)('Upload zip for browsers (%s)', browser => {
           cleanup([zipfile]);
         }
       }
-    });
+    }, 120000);
```
sufian-cf: Out of curiosity, why were these timeouts added?

src/index.ts
MildMax: Yes -- it has to do with how we determine whether or not the client has provided these args or if we should use the device emulation defaults. The problem with using integers as the default is that we aren't able to distinguish whether the client happened to provide the default values, or if they omitted height and width and assume the defaults.

__tests__/emulation.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+  describe.each(target_devices)(
+    'launchTest executes and returns result object when emulating device: %s',
+    device => {
+      let result: Awaited<ReturnType<typeof launchTest>>;
+      beforeAll(async () => {
+        let options = {
+          browser,
+          device: device,
+          url: 'https://www.example.com/',
```
sufian-cf, line 156: The prior tests use `url: '../tests/sandbox/index.html'`, which bypasses the network (more reliable tests). Is the network necessary here?

MildMax: yes, using `../tests/sandbox/index.html` causes the launchTest call to fail

these tests actually run the emulation all the way through, so the network calls are real

__tests__/artifacts.test.ts
MildMax: I think previously I had some trouble with timeouts on artifacts, but when I omit these now, they're fine -- I'll remove the timeout increase and use the default unless we hit the same problem in the future

src/index.ts
MildMax: applied the patch that removes treating the height/width optionally as a boolean and checks for undefined instead

sufian-cf (dismissed):

src/types.ts
```diff
@@ -656,4 +661,58 @@ export interface CLIOptions {
...
+  /**
+   * Indicates whether the device supports touch events.
+   * When true, enables touch event listeners and touch-specific behavior.
+   */
+  hasTouch: boolean;
+  /**
+   * The default browser engine to use for this device profile.
+   * Options: 'chromium', 'firefox', or 'webkit'
+   */
+  defaultBrowserType: 'chromium' | 'firefox' | 'webkit';
```
sergeychernyshev, line 719: I don't see this used anywhere and when device is ran with `--device`, but without specifying `--browser`, it always uses `chrome` instead of this property, e.g. for "iPhone 15", for example.

README.md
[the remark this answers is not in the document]
sergeychernyshev: Looks like this is implemented with "--device".

(@ryantownsend, it needs to be configured in the CLI params configuration and then the output is copied to README)

sergeychernyshev (dismissed):
Looks great - not sure if you want to include using default device as I commented below.

Also, I see the video is being weirdly cropped on smaller devices - not sure if this is a limitation of playwright.

src/types.ts
sergeychernyshev: As we discussed, let's merge it as is - I'll create a separate issue for this.

mjkozicki (approved):

sergeychernyshev (approved):
Releasing as is.

I created a separate ticket for defaultBroserType: #246

PR #140, opened 2026-02-19
switch from community package to official package for phosphor icons

joesanchezjr:
switched from community astro package to `@phosphor-icons/react`, using ssr version so we just get the svg - no js


joesanchezjr:
@Judyzc bundle is a bit larger for the worker, but not by much - but it's okay because what matters for this is what we're sending the client and that difference is negligible. I discussed this with @sergeychernyshev :)

PR #147, opened 2026-02-21
chore: switch to vitest

43081j:
i had this lying around locally from a while ago. i'm not sure what your plan is or if you even want to move to vitest, so i'll leave this as a draft for now.

IMO we should be using it as it is much faster and easier for contributors to pick up.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       HOME: /root
     container:
       # make sure to match this version with the one in package.json
-      image: mcr.microsoft.com/playwright:v1.57.0-noble
+      image: mcr.microsoft.com/playwright:v1.58.2-noble
 
     steps:
       - name: Checkout repository


sergeychernyshev (commented):
Looks like all reasonable changes - I tracked one problem to playwright upgrade - feel free to update it to 1.58.2, we are planning to do that regardless.

Also feel free to use HEADLESS and BROWSERS env variables to simplify testing:
https://github.com/cloudflare/telescope?tab=readme-ov-file#using-specific-browsers

I'll keep looking for other errors.

vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
```
sergeychernyshev:
```suggestion
  test: {
    testTimeout: 10000,
```
I got a few timeouts on some browsers, not sure if this is defaults difference between jest and vitest or just the performance difference, but let's bump this up to 10sec.

package.json
```diff
@@ -54,7 +53,7 @@
     "ejs": "^3.1.10",
     "ffmpeg": "^0.0.4",
     "msw": "^2.12.10",
-    "playwright": "^1.57.0",
-    "playwright-webkit": "^1.57.0"
+    "playwright": "^1.58.2",
+    "playwright-webkit": "^1.58.2"
```
sergeychernyshev: Looks like I forgot that you'd want to update the Docker image we run the tests in as well:
https://github.com/cloudflare/telescope/blob/main/.github/workflows/test.yml#L18

Issue #135, opened 2026-02-18
Parsed CLI JSON inputs are not validated


sergeychernyshev:
JSON inputs in CLI are not validated for conforming to their types.

Add `zod` library to properly validate the values.

dangayle:
I have a working branch with this fix almost ready to go, I'll submit a PR over the weekend or next week as soon as I am able.

Issue #149, opened 2026-02-24
Add command to view results of a previous test run


dangayle:
## Problem

After running a test, the only way to view results is to manually inspect files in the results directory. The `--html` and `--openHtml` flags only work at test time — there's no way to generate or open a report for a previously completed test.

## Expected behavior

A command like `npx telescope --open <testId>` (or similar) that generates and opens an HTML report from existing results.

## Current workaround

Manually open individual files (`metrics.json`, `screenshot.png`, `pageload.har`, etc.) from the results directory.

## Suggestion

Could reuse the existing HTML report generation logic but run it against a results directory instead of requiring a live test run.

PR #150, opened 2026-02-24
Validate CLI inputs at the boundary with Zod schemas

dangayle:
## Summary

We're using Commander's `.argParser()` to parse and validate all CLI inputs at the boundary, so only valid types flow into the application. By the time `normalizeCLIConfig` sees data, everything is already the correct type. No more `parseInt`/`parseFloat` calls or string-to-number coercion deep in the codebase.

We chose Zod for all inputs after finding several bugs in the existing rudimentary parsing:

- `parseInt(String(width)) || DEFAULT` silently swallows `0` (falsy) and falls back to the default, so `--width 0` gives you `1366`
- `parseFloat(cpuThrottle)` on a non-numeric string like `"fast"` produces `NaN` that propagates silently
- `flags.split(',')` on an empty string produces `['']` instead of `[]`, injecting an empty-string arg into Chromium's launch flags
- JSON options (`--cookies`, `--headers`, etc.) went through raw `JSON.parse` with no schema validation, so any valid JSON was accepted regardless of shape

Zod validates JSON options via `parseCLIOption()` (JSON parse + schema validation) and numeric options via `PositiveIntSchema`/`PositiveFloatSchema` in argParser callbacks. Zod exceptions are wrapped in Commander's `InvalidArgumentError` so invalid inputs print clean error messages instead of raw stack traces:

```
error: option '--height <int>' argument 'eight' is invalid.
       'eight' is not a valid --height value.
```

The `CLIOptions` type is updated to reflect reality: numeric fields are `number`, `flags` is `string[]`. Added 77 tests covering JSON syntax errors, schema validation per flag, backward compat, numeric schema edge cases, and the `parseWithSchema` error handling.

Closes https://github.com/cloudflare/telescope/issues/135
Relates to https://github.com/cloudflare/telescope/pull/33
Relates to https://github.com/cloudflare/telescope/issues/149

### DIFF
--- a/__tests__/validation.test.ts
+++ b/__tests__/validation.test.ts
@@ -16,11 +16,11 @@ import {
 import {
   parseCLIOption,
   parseUnknown,
+  parseWithSchema,
   formatZodError,
 } from '../src/validation.js';
 
 import { normalizeCLIConfig } from '../src/config.js';
-import { parseWithSchema } from '../src/validation.js';
 
 import type { CLIOptions } from '../src/types.js';
 


dangayle:
@sergeychernyshev It should be good to go now

Issue #151, opened 2026-02-24
Telescopetest-io: Add opencode AGENTS.md configs.


Judyzc:
## What is the problem your feature solves, or the need it fulfills?

It would be nice to have a specific Opencode agent setup for telescopetest-io. 

## Describe the solution you'd like

Just some setup for using Opencode with telescopetest-io, will probably be iteratively improved.

PR #152, opened 2026-02-24
Telescopetest-io: initial opencode setup

Judyzc:
Relates to #151 

*Edit - more info on how/why these files: 
- These files were generated initially with Opencode, but I then used them for some other coding and PR work in telescopetest-io and got Opencode to cut down on filler by basically telling it to not worry about being 'human-readable' and just focus on what is useful for it. 
    - Mainly wanted pointers about how to answer, to always be concise, to always use the internet (documentation) and re-read files before inferencing off of them, what I like for code style, code perferences, user preferences, specific env and configurations in the project, and any 'gatchas' (Opencode titled the section 'gatchas') that needed a bit more digging to find
    - Through trial and error I added some CLI commands that it could/couldn't run
    - Asked it to remove anything that was just code or structure regurgitated -> this is also good because now it's less git branch dependent -> when asking it to cut down, I found it useful to set limits (like 'give me one sentence' or 'give me bullet points each 10 words max')
  
In the future, I think adding more subagents rather than adding to the individual MD files is a good idea. Like a telescope.md file for knowing about the `telescope` CLI tool and the structure of tests in `/results` it outputs would probably be useful. 

I did read this recommended article and get a lot from it: https://addyosmani.com/blog/agents-md/

### DIFF
--- a/telescopetest-io/.opencode/agents/review.md
+++ b/telescopetest-io/.opencode/agents/review.md
@@ -19,7 +19,6 @@ CHECK FOR:
 - type for read models, interface for props
 - Explicit Prisma select
 - Null/undefined handling for D1 fields
-- R2: .head() before .get()
 - Zod validation on API inputs
 - API responses: { success, error? } with proper status
 - No user input in R2 keys without sanitization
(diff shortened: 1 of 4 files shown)


sergeychernyshev (commented):
I am trying to understand how to evaluate these.

Can you add some information in PR on how these were generated and how can they be used?

telescopetest-io/.opencode/agents/review.md
```diff
@@ -0,0 +1,33 @@
...
+
+CHECK FOR:
+
+- Scoped <style> per component
+- Server-side render first (flag unnecessary client scripts)
+- Named exports only (flag default exports)
+- type for read models, interface for props
+- Explicit Prisma select
+- Null/undefined handling for D1 fields
+- R2: .head() before .get()
```
sergeychernyshev: Is this a best practice of R2 of some kind? Feels like an extra unnecessary call to me without the context.

PR #158, opened 2026-02-25
Run telescope with Docker

thatzopoulos-cf:
Files Added
- Dockerfile - Multi-stage build using Node 24 with Playwright browser dependencies
- docker-compose.yml
- .dockerignore

Files Modified
- lib/browsers.js - Added Docker-specific flag (RUNNING_IN_DOCKER environment variable) to conditionally enable the following:
  - --no-sandbox and --disable-setuid-sandbox for Chromium browsers (Chrome won't start as root without this)
  - Enables ignoreHTTPSErrors: true for all browsers (handles Cloudflare WARP SSL interception. Without this, nothing works when Warp is enabled. Probably don't want to keep this)
  - Only affects Docker runs, not local development

Other Notes:
- Forced x86_64 emulation on Apple Silicon Macs (slower but required for the browsers we support)

Closes: https://github.com/cloudflare/telescope/issues/160

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM --platform=linux/amd64 node:24-bookworm
 
+ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
+
 # Install system dependencies for Playwright browsers and ffmpeg
 RUN apt-get update && apt-get install -y \
     ffmpeg \
@@ -40,8 +42,8 @@ RUN apt-get update && apt-get install -y \
 
 # Create non-root user for running Chrome securely (with sandbox enabled)
 RUN groupadd -r telescope && useradd -r -g telescope -G audio,video telescope \
-    && mkdir -p /home/telescope \
-    && chown -R telescope:telescope /home/telescope
+    && mkdir -p /home/telescope ${PLAYWRIGHT_BROWSERS_PATH} \
+    && chown -R telescope:telescope /home/telescope ${PLAYWRIGHT_BROWSERS_PATH}
 
 # Set working directory
 WORKDIR /app
@@ -64,7 +66,7 @@ RUN npm run build
 
 # Create results directory and set ownership for non-root user
 RUN mkdir -p /app/results /app/tmp /app/recordings \
-    && chown -R telescope:telescope /app
+    && chown -R telescope:telescope /app ${PLAYWRIGHT_BROWSERS_PATH}
 
 # Switch to non-root user for running Chrome with sandbox enabled
 USER telescope


Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+RUN npx playwright install chrome chrome-beta msedge
+
+# Copy the rest of the application
+COPY . .
+
+# Create results directory
+RUN mkdir -p /app/results /app/tmp /app/recordings
+
+# Default command - show help
+ENTRYPOINT ["node", "cli.js"]
```
Judyzc: Related to the bug above, if we're adding `RUN npm run build` I think we might also need this ENTRYPOINT to be `ENTRYPOINT ["node", "dist/src/cli.js"]`. But I might just be using this docker wrong. 

Kind of a separate question, but do we have an issue/docs for what this PR is doing and/or is that something we can add?

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      # Mount results directory to persist test results
+      - ./results:/app/results
+    # Required for Playwright browsers to run properly
+    ipc: host
+    # Security options needed for browser sandboxing
+    security_opt:
+      - seccomp=unconfined
+    environment:
+      # Run browsers in headless mode (required for Docker without display)
+      - CI=true
```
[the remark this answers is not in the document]
Judyzc: Actually I'm not sure if this is true I just checked my config.json from the test I ran with `docker compose run --rm telescope -u https://example.com -b chrome` and I did get "browser": chrome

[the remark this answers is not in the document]
thatzopoulos-cf: I think CI=true only impacts tests which I don't think we can run in the dockerimage with the new endpoint.

Judyzc (approved):

src/browsers.ts
```diff
@@ -61,20 +64,26 @@ class BrowserConfig {
...
+    ...(isDocker ? [] : ['--no-sandbox']),
   ];
 
   defaultBrowserOptions: Pick<
     BrowserConfigOptions,
     'headless' | 'viewport' | 'recordHar' | 'recordVideo'
   > = {
     headless,
+    // Ignore HTTPS errors in Docker - needed because Cloudflare WARP breaks SSL certificate validation
+    ...(isDocker && { ignoreHTTPSErrors: true }),
```
sergeychernyshev: Is there anything we can do to configure WARP to interact with Docker without this issue?

thatzopoulos-cf: Actually today it seems to be working today, maybe it wasn't warp. I will remove it

src/browsers.ts
```diff
@@ -61,20 +64,26 @@ class BrowserConfig {
     '--window-position="0,0"',
     '--window-size="1366,768"',
     '--remote-debugging-port=0',
+    // Required for running in Docker containers (Chrome won't start without these when running as root)
```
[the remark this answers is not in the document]
thatzopoulos-cf: With my latest changes you can run throttling with a command like this:
```docker compose run --user root telescope --connectionType 3G https://example.com```
Otherwise it will run as a non root user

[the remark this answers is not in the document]
Comment (sergeychernyshev) [reply src/browsers.ts]: Again, this might be more advanced setup, but can be the answer to limit privileges.

File: src/browsers.ts [hunk 5]
```diff
@@ -34,9 +34,12 @@ const headless: boolean = parseEnvBool(process.env.HEADLESS, CI);
 
 type BrowserConfigs = Record<BrowserName, BrowserConfigEntry>;
 
-// Docker environment flag - enables sandbox bypass and HTTPS error handling
+// Docker environment flag - used for HTTPS error handling
```
sergeychernyshev: I think you removed the wrong part of the comment here ;)

docker-compose.yml
```diff
@@ -12,5 +12,5 @@ services:
     environment:
       # Run browsers in headless mode (required for Docker without display)
       - HEADLESS=true
-      # Enable Docker-specific browser flags (--no-sandbox, ignoreHTTPSErrors)
+      # Enable Docker-specific flags (ignoreHTTPSErrors)
```
sergeychernyshev: and here

sergeychernyshev (changes requested):
Caught at the last moment, sorry.

sergeychernyshev (approved):

sergeychernyshev (approved):
Works on my home Docker setup no problem!

PR #164, opened 2026-03-02
Request priority in HAR file for Chromium based browsers

### LINKED ISSUE #16: Capture network/request priority in Chrome

ozcoder:
Adds `_priority` and `_initialPriority` to the HAR file when testing with a Chromium based browser.

Closes https://github.com/cloudflare/telescope/issues/16


sufian-cf:
The change itself looks reasonable, but I'm not able to verify it locally. How can I run telescope so that there are `_priority` and `_initialPriority` fields produced within the har?

MildMax (dismissed):
verified working order of the `initialPriority` , `priority`, and `resourceType` in the chrome browser

src/chromeRunner.ts
```diff
@@ -16,6 +16,37 @@ class ChromeRunner extends TestRunner {
...
+
+    let occasion = this;
+
+    // Just before request is sent
+    client.on('Network.requestWillBeSent', (params) => {
+      const { requestId, request, type } = params;
+      const fullURL = request.url + (request.urlFragment || '');
+      const resourceType = type || 'Other';
+
+      if (occasion.priorities[fullURL as keyof PriorityInfo]) {
```
sergeychernyshev: Can we use `requestId` as a key instead of request URL?

sergeychernyshev: Mapping by URL will break if there are multiple requests to the same URL.

sergeychernyshev (changes requested):
Looks like a good solution, we can make a couple improvements, see comments below.

Also, if you can make a fixture for testing the case with changing priority, it would be great. And maybe for testing duplicate requests.

src/chromeRunner.ts
ozcoder: I was using an array to get around that.
Now, _using a slightly more complicated way_, I can get the `requestId` from the CDP _universe_ to the Playwright universe.

src/chromeRunner.ts
```diff
@@ -16,6 +16,37 @@ class ChromeRunner extends TestRunner {
...
+        occasion.priorities[fullURL].push({
+          requestId: requestId,
+          initialPriority: request.initialPriority,
+          resourceType: resourceType
```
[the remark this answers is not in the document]
ozcoder: Done.

packages/telescope/src/chromeRunner.ts
```diff
@@ -16,6 +17,34 @@ class ChromeRunner extends TestRunner {
...
+      const { requestId, newPriority } = params;
+
+      occasion.newPriorities[requestId] = newPriority;
+    });
+
+    client.on('Network.responseReceived', (params) => {
+      const { requestId, response } = params;
+
+      if (response.timing) {
+        const hashString = crypto.createHash('sha256').update(response.url + response.timing.sendStart).digest('hex');
```
sufian-cf: Here we hash `response.url + response.timing.sendStart`

But when we look the hash up (in `src/testRunner.ts`), we use `request.url + request.timing.requestStart`

Is this difference intentional?

packages/telescope/src/chromeRunner.ts
```diff
@@ -16,6 +17,34 @@ class ChromeRunner extends TestRunner {
   async createPage(browser: BrowserContext): Promise<Page> {
     const page = browser.pages()[0];
     const client: CDPSession = await page.context().newCDPSession(page);
+    await client.send('Network.enable');
+
+    let occasion = this;
```
sufian-cf: Nit: this closure isn't needed -- `this` will still be bound to the same object in the event handlers below, since they're arrow functions.

packages/telescope/src/chromeRunner.ts
```diff
@@ -16,6 +17,34 @@ class ChromeRunner extends TestRunner {
   async createPage(browser: BrowserContext): Promise<Page> {
     const page = browser.pages()[0];
     const client: CDPSession = await page.context().newCDPSession(page);
+    await client.send('Network.enable');
```
sufian-cf, line 21: Is it possible `Network.requestWillBeSent` / other events below will fire _before_ this `client.send('Network.enable')`?

sufian-cf:
Hey @ozcoder, my apologies for the delay. I've taken a look at this approach and have manually verified that it _sometimes_ works, but not for all requests. I've left a few comments, and overall am a bit concerned with the approach.

That being said, I think this has surfaced a real bug in our existing code. If the page load triggers multiple requests with the same URL, we're going to only be updating performance information on the first one, since we look up the corresponding entry via `findIndex` on just the URL.

I've just created https://github.com/cloudflare/telescope/issues/243 and will follow up with a fix that introduces a unique id per request we can use to match things.

sufian-cf:
Ok, I have a PR for the fix in #243 which should give each request a unique identifier that can be used to cross-reference: https://github.com/cloudflare/telescope/pull/244

Once that mechanism is used, and there's some automated test coverage verifying the presence of the priority fields in the HAR, this PR should be good to get in

packages/telescope/src/chromeRunner.ts
ozcoder: They are the same value. It's because we are dealing with the **CDP** _universe_ in one place and the **Playwright** _universe_ in the other place. I had to find a _stable_ attribute that was _known_ to the request and the response and would be different when the same URL is requested more than once.

packages/telescope/src/chromeRunner.ts
ozcoder: Ah, yes. Old habit.

packages/telescope/src/chromeRunner.ts
ozcoder: Shouldn't happen since even `preparePage` hasn't been called until the end of this method.

packages/telescope/src/chromeRunner.ts
sergeychernyshev: @ozcoder, I merged #244 so you should be able to use `telescopeId` for reconciling HAR and network results.

ozcoder: OK

packages/telescope/src/chromeRunner.ts
ozcoder: removed.

ozcoder:
> Hey @ozcoder, my apologies for the delay. I've taken a look at this approach and have manually verified that it _sometimes_ works, but not for all requests. I've left a few comments, and overall am a bit concerned with the approach.
> 
> That being said, I think this has surfaced a real bug in our existing code. If the page load triggers multiple requests with the same URL, we're going to only be updating performance information on the first one, since we look up the corresponding entry via `findIndex` on just the URL.
> 
> I've just created #243 and will follow up with a fix that introduces a unique id per request we can use to match things.

Changed to use `telescopeId` .

ozcoder:
@sufian-cf @sergeychernyshev I have changed the code to use `telescopeId` and added a test.
`npm run test priority` to run it. I'm not sure what else to add to this change.

MildMax:
@ozcoder apologies, closed by accident

ozcoder:
@MildMax With the test server, I have added the ability to _delay_ the responses to _simulate_ processing time. I also added a compression option, which adds some _jitter_ to the processing time and _might_ be useful for other tests. To see why the **GitHub** CI might not be happy, I now show launchTest error if not successful.

sergeychernyshev:
@ozcoder looks like one of the new tests is failing

ozcoder:
> @ozcoder looks like one of the new tests is failing

Oh, it's because the test is **chrome** specific which isn't installed in the CI.
Is **chromium** available in the workflows? I would need to add `chromium` to `browsers.ts` and `types.ts`

sergeychernyshev:
> > @ozcoder looks like one of the new tests is failing
> 
> Oh, it's because the test is **chrome** specific which isn't installed in the CI. Is **chromium** available in the workflows? I would need to add `chromium` to `browsers.ts` and `types.ts`

Ahh. Yeah, that makes sense. Can you limit the test to Chromium then? e.g. skip it if the browser is not Chromium so it runs locally, but doesn't block CI.

We should add other browsers to CI too - just haven't had time to work on that yet.

ozcoder:
@sergeychernyshev The `playwright:v1.59.1-noble` docker image that the CI uses has **chromium**, **webkit**, **firefox** and  **chromium_headless_shell** browsers installed. I changed the test to use **chromium** and added it to `BrowserConfig`, `BrowserName` and `BrowserChannel`. The `priority` test should now be able to run in the CI. When you have time can you please let it run in the CI so we all can see it run.

ozcoder:
Fixed failing envVars test.

MildMax (approved):
Tested the changes against the provided tests/priority/index.html as well as wikipedia pages successfully -- priority and initialPriority are in the chrome HAR files and resourceType

sergeychernyshev:
@ozcoder please remind me why you added `chromium` as an option? did `chrome` not work for some reason?

ozcoder:
> @ozcoder please remind me why you added `chromium` as an option? did `chrome` not work for some reason?

The docker image the CI uses doesn't have Chrome (maybe for legal reasons), but does have Chromium, the engine that powers Chrome.

sergeychernyshev (approved):
OK, confirmed that CI runs it in chromium specifically.

sergeychernyshev:
Thank you, @ozcoder - it was a long battle, but we are finally there!

We will release new version of the package later this week.

sergeychernyshev:
🎆 🍾 🇦🇺

Issue #71, opened 2025-12-15
Timeline + Waterfall viewer


slightlyoff:
A small web component that consumes the timeline images and `resources.json` for displaying scrolling timelines would be useful for understanding results.

I'm happy to contribute one if there's interest.

sergeychernyshev:
I am still working on #177, but the goal is to implement a self-contained, framework-free web component (`<waterfall-chart>`) that renders HTTP Archive (HAR) waterfall charts.

All styling is in a standalone `waterfall.css` — the chart renders correctly with or without JavaScript.

From dashboards like @pmeenan's [waterfall-tools](https://github.com/pmeenan/waterfall-tools) to blog post illustrations, we want to make this component usable across different apps (including our own `telescope-web`, obviously).

## Rendering Modes

1. **Static (CSS-only)** — Pre-render HTML at build time via `renderToHTML(har)`. No JS bundle required; the chart is fully functional from HTML + CSS alone. _This might be useful for rendering images or for non-interactive UIs of some kind. Mostly exists to support progressive enhancement mode._
2. **Progressive Enhancement** — Same pre-rendered HTML, but the JS bundle loads lazily. The element upgrades in place, adding interactivity without re-rendering. _This is a default mode of operation with server-side rendered HTML that renders firs then converts to interactive._
4. **Fully Dynamic (JS-driven)** — Supply HAR data via the `src` attribute (fetched from a URL) or the `.har` JS property. The element builds its own DOM from scratch. Reassigning either value triggers a full re-render. _This is great for client-side only implementations that just want to render a HAR file._

## Interactive Features (JS mode)

- **Resource-type filter chips** — Toggle visibility of rows by type (`html`, `js`, `css`, `image`, `font`, `video`, `other`). Multiple types can be active simultaneously.
- **Column toggle** — Expand/collapse detail columns (Method, Protocol, Status, Type, Size, Duration).
- **Timeline scrubber** — Hover over the waterfall to see a vertical time indicator (ms). Snaps to event lines (DCL, Load, LCP) within 8 px and displays their full label.
- **Detail panel** — Click any row to open a panel with request/response timings, headers, and general info. Click again or use the close button to dismiss. In the future releases we'll probably create a few variations of this and make and allow replacing it with callbacks so developers can implement their own detail panels.

## Theming

[...]

PR #177, opened 2026-03-05
feat(waterfall): HAR waterfall chart web component

sergeychernyshev:
## Summary

Introduces `@cloudflare/waterfall` — a standalone, framework-free web component that renders an HTTP Archive (HAR) waterfall chart.

Contributes to #71

---

### Web component (`src/`)

- **`<waterfall-chart>`** custom element with three usage modes:
  - **Static** — pure HTML + CSS, zero JavaScript, pre-rendered via `renderToHTML()`
  - **Progressive enhancement** — pre-rendered children upgraded in place when the JS bundle loads, wiring up interactivity without re-rendering
  - **Fully dynamic** — element builds its own DOM from a HAR supplied via the `src` attribute (fetched) or the `.har` JS property
- **`src/waterfall.css`** — standalone stylesheet; link in `<head>` with or without the JS bundle
- **`src/render.ts`** — pure `renderToHTML(har)` function, runs in Node.js or the browser, no DOM dependency
- **`src/waterfall-chart.ts`** — custom element implementation: filter chips by resource type and connection phase, row click → detail panel, column toggle, scrubber, ResizeObserver-accurate event lines (DCL, Load, LCP)
- **`src/index.ts`** — barrel export (`renderToHTML`, `WaterfallChart`)
- Supporting modules: `har.ts` (HAR 1.2 types), `config.ts`, `helpers.ts`, `formatters.ts`

### Demo site (`public/`)

Served with Vite (`npm run dev`), with `public/` as the web root:

| Page | Description |
|------|-------------|
| `index.html` | Progressive enhancement — pre-rendered chart, click to activate JS |
| `interactive.html` | Fully dynamic — load any local HAR file |
| `src-attr.html` | Load HAR via the `src` attribute |

Component source files are kept in `src/` but exposed to the demo at clean URLs via a Vite plugin:

| URL | Source |
|-----|--------|
| `/waterfall/waterfall.css` | `src/waterfall.css` |
| `/waterfall/index.ts` | `src/index.ts` |

Shared demo assets (`demo.css`, `theme.js`, `progressive.js`, `interactive.js`) live alongside the HTML in `public/`.

### Tooling

- **`vite.config.ts`** — Vite dev server with `root: public`, custom `srcAliasPlugin` mapping `/waterfall/*` URLs to `src/`
- **`scripts/gen-demo.js`** — regenerates pre-rendered HTML in `index.html` and `interactive.html` from a HAR file; run after `npm run build`
- **`tsconfig.json`** — targets ES2020, `moduleResolution: bundler`, strict mode
- **Playwright + Vitest** test suite covering bar rendering, metric filters, overlay positioning, theme toggle, column toggle

### Package

- Name: `@cloudflare/waterfall`
- Exports: `dist/index.js` (ESM) + `dist/index.d.ts`
- Built with `tsc`; `npm run dev` starts the Vite demo server


packages/waterfall/src/render.ts
```diff
@@ -0,0 +1,383 @@
...
+      </div>
+      <div class="wf-events-overlay" aria-hidden="true">
+        ${eventLinesHTML}
+      </div>
+    </div>
+  </div>
+  <ol class="wf-list" aria-label="Network requests">
+    ${rowsHTML}
+  </ol>
+</div>
```
ryantownsend: Quick note so it doesn't get lost:
 
* We probably shouldn't be hiding all these from accessibility tooling.
* We should question whether this should be a `<table>` for accessibility (or at least use aria `role`/`scope` attributes (afaik the limitations/impacts of table layout and parsing can be overridden in CSS these days)

sergeychernyshev: If you have any good article on how to do all that, or an AI skill I can use for that, it would be awesome.

sergeychernyshev: I don't like div-itis either and would rather use semantic elements, especially if they come with built-in browser behaviors like that.

sergeychernyshev: I added a separate ticket for this work: https://github.com/cloudflare/telescope/issues/274

packages/waterfall/__tests__/helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export function createServer(): Promise<{ url: string; close: () => void }> {
+  return new Promise((resolve) => {
+    const server = http.createServer((req, res) => {
+      const filePath = path.join(
+        ROOT,
+        req.url === '/' ? '/static.html' : (req.url ?? ''),
+      );
+      try {
+        const data = fs.readFileSync(filePath);
```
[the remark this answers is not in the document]
sergeychernyshev: This is a unit test.

packages/waterfall/src/waterfall-chart.ts
```diff
@@ -0,0 +1,1364 @@
...
+    this._rulerEl.innerHTML = '';
+    this._gridOverlayEl.innerHTML = '';
+    this._overlayEl.innerHTML = '';
+    this._overlayEl.appendChild(this._scrubberEl);
+    this._errorEl.hidden = true;
+    this._errorEl.textContent = '';
+    // Remove any open detail panels
+    this.querySelectorAll('.wf-panel').forEach((p) => p.remove());
+    this._showLoading(true);
+  }
```
sufian-cf: Looks like this is never called?

packages/waterfall/src/waterfall-chart.ts
```diff
@@ -0,0 +1,1364 @@
...
+      this._loadHarData(this._harData);
+    } else if (this.hasAttribute('src')) {
+      this._buildDOM();
+      this._fetchAndRender(this.getAttribute('src')!);
+    } else if (this.querySelector('.wf-list')) {
+      // Pre-rendered static HTML children detected — adopt and wire up
+      this._adoptDOM();
+    }
+    // else: empty element, nothing to do until src/har are set
+  }
```
sufian-cf, line 142: I'm pretty sure this'll cause problems in many frameworks -- connectedCallback is called when the element is attached to the DOM, but it could be temporarily detached and then reattached without recreating.

We should probably have a `this._initialized` boolean which causes the initialization to occur once on `connectedCallback`.

packages/waterfall/src/render.ts
```diff
@@ -0,0 +1,314 @@
...
+      ? [`data-transfer-size="${data['transfer-size']}"`]
+      : []),
+  ].join(' ');
+
+  return `<li class="${rowClassStr}" ${dataAttrs}>
+      <span class="wf-cell wf-cell--idx">${visIdx}</span>
+      <span class="wf-cell wf-cell--url" title="${esc(entry.request.url)}"><span class="wf-url-domain">${esc(domain)}</span>${pathCell}</span>
+      <span class="wf-cell wf-cell--info">${esc(entry.request.method)}</span>
+      <span class="wf-cell wf-cell--info">${esc(entry.request.httpVersion)}</span>
+      <span class="wf-cell wf-cell--info wf-cell--stat ${statusCls}">${status}</span>
```
sufian-cf: `${status}` should be escaped -- it's potentially untrusted user input

sufian-cf (changes requested):
Overall looks good. We'll likely run into problems with React/other frameworks, but this seems like a good start and can be merged almost as-is so we can move forward.

The only blocking/concerning issue I see is the non-escaping of the `status` value (from `entry.response.status` in `render.ts`'s `renderRow()` -- I'm 90% certain that opens up a potential XSS from specially-crafted .har files that contain HTML in the status field.

packages/waterfall/src/render.ts
sergeychernyshev: I updated the code to use `html` template literals to reduce the chance of this happening again.

packages/waterfall/src/waterfall-chart.ts
sergeychernyshev: I agree - this is a challenge that we'll need to overcome when we integrate with frameworks that keep their DOM nodes closer to the chest.

packages/waterfall/src/waterfall-chart.ts
sergeychernyshev: Removed it and scoured for other unused code which found a minor duplication in tests which I also refactored.

.github/workflows/deploy.yml
```diff
@@ -17,10 +17,10 @@ jobs:
...
       - name: checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: setup Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
           node-version: 'lts/krypton'
           cache: 'npm'
```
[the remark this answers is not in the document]
sergeychernyshev: That's incorrect, `lts/krypton` is node 24, not node 6.

sufian-cf (approved):
Looks good, thank you!

PR #203, opened 2026-03-16
Telescopetest-io: Add Initial Filmstrip & Video tab

### LINKED ISSUE #189: Test Results - Filmstrip & Video
* All filmstrip  images into a horizontal timeline. Include the time from start in milliseconds below. Ideally this would be linked to the video and progress in sync.
* FCP, LCP frames need to be highlighted by border color.
* Layout shifted frames need highlight and overlay.
* Video section - ideally this could be synced to the filmstrip timestamps.

Judyzc:
Adds simple filmstrip & video tab to telescopetest.io/results. Closes #189 
- After talking with Sergey, the filmstrip should line up with the waterfall, NOT the video 
- Will think on how to link filmstrip to waterfall after waterfall component gets reworked/wrapped into telescopetest-io
- Filmstrip timestamp end duration doesn't line up fully with video, as in the filmstrip usually goes on for much longer than the video. This could be some kind of issue with how filmstrip images are generated/numbered in telescope. This filmstrip also starts from 0s, which is different from telescope's `--html` option output which starts from 0.1s or 1s. 
- Changed `[filename].ts` to `[...filename].ts` to allow this endpoint to get one dir down nested resources, using Astro's dynamic API routing: https://docs.astro.build/en/guides/routing/#on-demand-dynamic-routes
    - Tested that this didn't break serving other resources 
- Checked on staging

### DIFF
--- a/telescopetest-io/src/components/FilmstripVideo.astro
+++ b/telescopetest-io/src/components/FilmstripVideo.astro
@@ -59,7 +59,7 @@ try {
     })
     .sort((a, b) => a.frameNum - b.frameNum);
   filmstripFrames = frames.map(frame => {
-    // try to calculate filmstrip timestamps, assume frame # happesn at # second. might be wrong in telescope because ffmpeg in telescope extracts frames from video ?
+    // try to calculate filmstrip timestamps, assumes frame # happens at # second.
     // to get more precise framerates, also store video duration.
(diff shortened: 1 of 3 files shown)


MildMax (dismissed):
Verified working order of the filmstrip & video tab in local env; follow up issue created from prior review

PR #207, opened 2026-03-17
Telescopetest-io: fix 'UNSAFE' rated content from being available by endpoint

### LINKED ISSUE #208: Telescopetest-io: BUG with AI content rating
I realized there was a bug with the AI content rating where content_rating='UNSAFE' test content can still be reached by endpoint.

- For example, test_id='2026_02_20_16_56_43_5b5892e4-11f5-4475-8c34-6daf8d0f96dd' is marked unsafe in production DB. Going to https://telescopetest.io/results/2026_02_20_16_56_43_5b5892e4-11f5-4475-8c34-6daf8d0f96dd correctly shows that this result is unsafe and tells the user they will need to run telescopetest-io locally. I think it makes sense that this badge is shown. 
- However, by going to the api/tests/[testId]/screenshot.png endpoint directly, you can still see the unsafe content. Example: [image of 'violent movie' youtube screenshot](https://telescopetest.io/api/tests/2026_02_20_16_56_43_5b5892e4-11f5-4475-8c34-6daf8d0f96dd/screenshot.png)

Judyzc:
Fixes #208 

This PR just tries to fix that: 
- Changed `[filename].ts` to check content rating first before serving anything from R2. 
- Added `aiEnabled` param to `getTestById` prisma function so only serve results in `[testId].astro` if safe.
- Tested that this bug is fixed on staging.

### DIFF
--- a/telescopetest-io/src/lib/repositories/testRepository.ts
+++ b/telescopetest-io/src/lib/repositories/testRepository.ts
@@ -67,7 +67,7 @@ export async function getTestById(
       content_rating: true,
     },
   });
-  return row;
+  return row ?? null;
 }
 
 /**


sergeychernyshev:
✨ AI-assisted

Issue #154, opened 2026-02-25
Telescopetest-io manual ZIP upload error


Judyzc:
## Describe the bug

A test result folder that has a 'config.json' file that I manually ZIP (in finder/files) and then try to upload on telescopetest.io/upload returns an 'Error: No config.json file found in the ZIP archive' error. 

## Telescope info

Please include the following information about your environment:

**Telescope version**: 49344f76ee9977b53c1b7a329b7922efdaf7d097
**Node version**: v24.13.1
**Operating system version**: macOS 26.3 

## Steps to reproduce

I had an unzipped telescope `results/` test that I know had a `config.json` file that I then compressed and tried to upload.

## Expected results

Proper upload

## Observed results

I kept running into an error on the upload page and the test never appeared in telescopetest.io/results.

PR #210, opened 2026-03-18
Telescopetest-io: fix error with manual ZIP uploads by parsing filepaths 

Judyzc:
Addresses #154 

- The error seems to be that the filepath for manually ZIPPED files includes the folder name, while the filepath for telescope agent `--zip` outputs do not. 
- In `upload.ts` for the POST endpoint, I added a helper function `normalizeZipFiles()` that strips directory prefixes and filters unwanted files from upload. Prefixes are based on `config.json` file, which should always be present in a test. I'm also assuming `config.json` is always on the top level. 
- Tested on staging

### DIFF
--- a/telescopetest-io/.opencode/agents/ttio.md
+++ b/telescopetest-io/.opencode/agents/ttio.md
@@ -1,6 +1,6 @@
 ---
 description: Assistant for the telescopetest-io project. Use for general development, feature work, debugging, and questions about this codebase.
-model: anthropic/claude-sonnet-4-6
+model: anthropic/claude-sonnet-4-5-20250929
 mode: primary
 permission:
   edit: ask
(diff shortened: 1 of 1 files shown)


telescopetest-io/src/pages/api/upload.ts
```diff
@@ -107,9 +118,11 @@ export const POST: APIRoute = async (context: APIContext) => {
         },
       );
     }
-    // Confirm the config file exists
-    const configFile = `config.json`;
-    if (!files.includes(configFile)) {
+    // Find config.json to determine the path prefix
+    const configPath = files.find(
+      f => f.endsWith('/config.json') || f === 'config.json',
```
sergeychernyshev: let's use `path.basename('config.json')` to determine the config file entry

Issue #117, opened 2026-02-07
telescopetest.io homepage does not support light theme


sergeychernyshev:
Telescopetest.io homepage is not responsive to `prefers-color-theme` property and does not support light theme.

Update the theme to make it responsive.

Also, remove the purple AI stench it has all over it.

PR #213, opened 2026-03-19
Telescopetest-io: add nav bar to home page 

Judyzc:
- Just adds `TopNav.astro` component to the <body> of `index.astro`, and that's it. Light/dark mode toggles on home page don't change anything on the home page.
- Adding light/dark mode toggle to home page already tracked with separate ticket: https://github.com/cloudflare/telescope/issues/117
- Made some changes on `Page.astro` and `TopNav.astro` to make sure nav bar stays with screen as scroll down. 
- Tested on staging

### DIFF
--- a/telescopetest-io/src/components/TopNav.astro
+++ b/telescopetest-io/src/components/TopNav.astro
@@ -1,4 +1,8 @@
 ---
+interface Props {
+  hideThemeSwitcher?: boolean;
+}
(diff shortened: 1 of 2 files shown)


sergeychernyshev (commented):
Just hide the theme switcher on homepage and you should be good to go.

sergeychernyshev:
We might need to wait till #202 and #203 are merged in though.

PR #219, opened 2026-03-23
Accessibility testing with aXe

### LINKED ISSUE #60: Accessibility tests with Axe runner for Playwright
## What is the problem your feature solves, or the need it fulfills?

A clear and concise description of why this feature should be added. What is the problem? Who is
this for?

I looked at this project and as it is Playwright based, I thought it would have been great to apart auditing performance, it cold also audit Accessibility using Axe runner for Playwright.

## Describe the solution you'd like

What do you propose to resolve the problem or fulfill the need above? How would you like it to
work?

I think the option to run the test with the Axe runner as shown here https://playwright.dev/docs/accessibility-testing could be an advantage. Maybe an optional flag like `--a11y` and if so the test will be run for the website and the results will be reported :)

## Describe alternatives you've considered

What other solutions, features, or workarounds have you considered that might also solve the issue?
What are the tradeoffs for these alternatives compared to what you're proposing?

Basically doing the A11Y tests separately with https://playwright.dev/docs/accessibility-testing

## Additional context

This could include references to documentation or papers, prior art, screenshots, or benchmark
results.

Sergey showed this project during the recent GDE Web Perf Meeting and I really liked it!

ozcoder:
Added the `--a11y` command line option to run an accessibility test on the webpage. It produces a JSON report file called `a11y.json`.

The options for the command line are in JSON format. Example :-

`--a11y '{"tags": ["wcag2aa"], "disable": ["color-contrast"]}'`

  * **tags** is the name of a group of rules, usually a Standard name. Given as an array of strings.
  * **disable** is to disable a particular rule that you do not want in the report. Given as an array of strings.
  * **exclude** is to exclude a particular HTML tag element from being checked. Given as an array of CSS Selector strings.

See https://github.com/dequelabs/axe-core/blob/HEAD/doc/rule-descriptions.md  for more information.

Closes https://github.com/cloudflare/telescope/issues/60


mjkozicki:
We are good to go with this. @sergeychernyshev will need to review.

@ozcoder

MildMax:
Hi @ozcoder , thanks a ton for putting this together (and sorry for the long delay on getting back to you here).

We agree accessibility testing is important, and integrating aXe with Playwright is a reasonable approach. That said, we’re not in a position right now to take on and maintain new Telescope features that aren't performance-related (new CLI surface area, dependency, docs/support), so we’re going to leave this PR parked for now (no ETA), since we can’t commit to maintaining this feature at the moment.

If you’d like to keep using this in your fork, please do. For folks who want aXe results today, the best path for now is to run aXe alongside Telescope in their own Playwright/a11y harness, rather than as a built-in Telescope feature.

Thanks again for the work and for your patience.

PR #223, opened 2026-03-24
Upgrade to Astro 6 and @astrojs/cloudflare v13

joesanchezjr:
Upgrades the Astro framework and the Cloudflare adapter to their latest major versions - which brings the local dev environment into full parity with the production Workers runtime

### What changed

**Dependencies**

- `astro`: `^5.16.15` → `^6.0.8`
- `@astrojs/cloudflare`: `^12.6.12` → `^13.1.3`
- `@astrojs/react`: `^4.4.2` → `^5.0.1`
- `cross-env` added to `devDependencies` for cross-platform script compatibility

**Cloudflare env access** (`Astro.locals.runtime` API removed in v13)

All bindings previously accessed via `Astro.locals.runtime.env` are now imported directly:

```ts
// before
const env = context.locals.runtime.env;

// after
import { env } from 'cloudflare:workers';
```

**Local dev environment**

The old `platformProxy` config in `astro.config.mjs` is replaced by setting `CLOUDFLARE_ENV=development` when starting the dev server. This tells the Cloudflare Vite plugin to use the `env.development` block from `wrangler.jsonc`, giving `astro dev` access to the real local D1 database and R2 bindings.

```json
"dev": "cross-env CLOUDFLARE_ENV=development astro dev"
```

**Prisma + workerd compatibility**

Replaced `vite.ssr.external` and `optimizeDeps.include` with a targeted Vite plugin using the `configEnvironment` hook — the pattern recommended in the Astro Cloudflare upgrade docs (https://docs.astro.build/en/guides/integrations-guide/cloudflare/#some-dependencies-might-need-to-be-pre-compiled):

```js
{
  name: 'pre-compile-deps',
  configEnvironment(name) {
    if (name !== 'client') {
      return {
        optimizeDeps: {
          exclude: ['@prisma/client/runtime/wasm-compiler-edge'],
        },
      };
    }
  },
}
```

Two issues required this:

1. `vite.ssr.external` was dead config. The adapter's internal `cf-externals` Vite plugin sets `ssr.noExternal = true` and clears `ssr.external` post-config. Node.js built-ins are handled by the `nodejs_compat` compatibility flag in `wrangler.jsonc` instead.

2. `wasm-compiler-edge` must not be bundled by esbuild. The generated Prisma client at `generated/prisma/internal/class.ts` loads its WASM binary via `import("./query_compiler_fast_bg.wasm?module")`. The `?module` suffix is a workerd-native syntax that tells the runtime to instantiate the file as a `WebAssembly.Module`. esbuild does not understand this syntax — when it tries to pre-bundle `wasm-compiler-edge`, it corrupts the import and the WASM decoder throws `TypeError: Failed to decode input`. Excluding it from the optimizer lets the `?module` import reach workerd intact.

**`wrangler.jsonc` entrypoint**

Updated `main` to the new v13 unified entrypoint:

```
"main": "@astrojs/cloudflare/entrypoints/server"
```

**CSS extraction**

Moved the large `<style is:global>` block out of `Layout.astro` into `src/styles/global.css`, imported via ESM. No visual changes.

**Dev setup script**

Removed the `wrangler r2 bucket create` step from `dev-setup.sh`. Wrangler creates the local R2 bucket directory automatically on first use based on `wrangler.jsonc`, so the explicit creation step was unnecessary for local development.

Closes https://github.com/cloudflare/telescope/issues/220

### DIFF
--- a/telescopetest-io/README.md
+++ b/telescopetest-io/README.md
@@ -27,9 +27,7 @@ You should now be able to use the application and run `npm run studio` to view l
 
 ## Running Locally
 
-Make sure you've followed all steps in Project Setup above.
-
-Then, you can run `npm run build` and then `npm run dev` to view the site with Astro's hot reload (instantly reflect changes) using the adapter for Cloudflare. Alternatively, you can run `npm run preview` to see Astro with Workers together in one step, but there's no hot reload.
+Run `npm run dev` to view the site with Astro's hot reload (instantly reflect changes) using the adapter for Cloudflare.
 
 ## DB Migrations
 
@@ -52,7 +50,7 @@ npx prisma migrate diff \
 This should fill your created file with the raw SQLite for your changes.
 
 4. Run `npm run generate` to regenerate a Prisma Client that reflects your new changes in `schema.prisma`.
-5. Run `npm run migrate:local` to apply this new migration to your local database.
+5. Run `npm run migrate:development` to apply this new migration to your local database.
 
 ### Note about Workers AI (AI content review)
 
--- a/telescopetest-io/package.json
+++ b/telescopetest-io/package.json
@@ -8,7 +8,7 @@
     "dev:clean": "rm -rf .wrangler .env dist node_modules",
     "cf-typegen": "wrangler types",
     "build:development": "cross-env CLOUDFLARE_ENV=development astro build",
-    "build:staging": "cross-env CLOUDFLARE_ENV=production astro build",
+    "build:staging": "cross-env CLOUDFLARE_ENV=staging astro build",
     "preview": "astro preview",
     "migrate:development": "npx wrangler d1 migrations apply telescope-db-development --local --env development",
     "migrate:staging": "npx wrangler d1 migrations apply telescope-db-staging --remote --env staging",


Judyzc (changes requested):
Just a few changes, I just tested by deleting my local `.wrangler` and `.node-modules` and tried to install/run the project from scratch. 

1. From `npm install`, I get the usual npm audit output 
```
up to date, audited 423 packages in 2s

156 packages are looking for funding
  run `npm fund` for details

11 vulnerabilities (5 moderate, 6 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
```
But when I run `npm audit fix --force` (which I don't know if you're supposed to do but I like to, and so do some of my friends), prisma version in `package.json` gets bumped down to version 6 from version 7. This causes some issues in `npm run dev:setup` and I'd imagine possibly other places. Maybe we can put some note to not run `npm audit fix --force`? Or AI tells me there's some "override" section we can add to the package.json to force prisma version 7, like this: 

```
{
  "name": "telescopetest-io",
  "version": "0.0.1",
  "overrides": {
    "prisma": "^7.3.0",
    "@prisma/client": "^7.3.0",
    "@prisma/adapter-d1": "^7.3.0"
  },
  "engines": {
    "node": ">=24",
    "npm": ">=11.9.0"
  },
  ...
}
```


2. When I run `npm run build` and then `npm run preview`, I just get a blank page with "[Object object]" from the page. 

Image

I'm not sure in `npm run preview` is something we even need anymore honestly since `npm run build` and then `npm run dev` now uses bindings (?) directly ?

telescopetest-io/src/lib/ai/ai-content-rater.ts
```diff
@@ -1,6 +1,5 @@
 import { ContentRating } from '@/lib/types/tests';
 
-type Ai = import('@cloudflare/workers-types').Ai;
 
 // helper to extract text from metrics.json
```
sergeychernyshev: Nit: removing the `type Ai` line leaves a double blank line here. Consider removing one of them to keep the file tidy.

Issue #209, opened 2026-03-17
Telescopetest-io: Use Cloudflare workflows to remove AI content rating timeout and bottleneck in serving 'SAFE' assets


Judyzc:
## Problem
1) AI content rating in telescopetest-io runs using `ctx.waitUntil()` which has a 30-second timeout. If AI models take longer, the rating fails silently and tests stay stuck as UNKNOWN/IN_PROGRESS.
2) Protecting each asset in R2 (accessed through the endpoint api/tests/{testId}/{filename}) puts a lot of strain on D1 (to fetch the content_rating of one test_id multiple times). This sometimes results in 500 errors (#221), and we need a more permanent solution than #224. 

### Problem 1 Solution
Use Cloudflare Workflows to run AI rating without timeout limits. 

#### What to Change
1. Create new workflow file `src/workflows/ai-content-rating.ts`
   - 4 steps: mark in-progress → fetch R2 files → run AI → update DB
2. Update `wrangler.jsonc`
   - Add workflow binding for dev/staging/prod
3. Update `src/pages/api/upload.ts`
   - Remove `ctx.waitUntil()` block
   - Replace with `env.AI_RATING_WORKFLOW.create({ params: { testId, url } })`
4. Update `worker-configuration.d.ts`
   - Add `AI_RATING_WORKFLOW: Workflow` to Env interface
- No changes needed: Database schema, Repository functions, Detail page, Rating API endpoint, Frontend polling logic

#### How It Works
1. User uploads → API saves test → triggers workflow → returns immediately
2. Workflow runs in background (no timeout)
3. Frontend polls `/api/tests/[testId]/rating` until done
4. When workflow finishes, frontend detects rating changed

### Problem 2 Solution 
- Ideal implementation we should serve test results directly from a public R2 bucket though CDN rather than using a worker for loading them from R2.
- Which means we do need to implement a private upload bucket where assets go before they are rated (if AI rating is enabled).
- Might be a good task to implement using those workflows.
- Most docs suggest standalone custom domain, but I can imagine we could map a subpath of the main domain to that bucket.
- If we set up separate buckets, we'd be able to do #216 much more easily

needlecodev:
This issue describes two main problems: a timeout issue with AI content rating due to `ctx.waitUntil()` and performance bottlenecks when serving assets.

[...]

PR #234, opened 2026-03-27
Add workflows + second public R2 bucket

Judyzc:
Addresses (fixes) #209. Tested on staging.

### High-level explanation of changes:  
- 1st Problem: AI content rating could time out after 30 seconds because Cloudflare Workers can only run a request for 30 seconds. This can leave a test in "IN_PROGRESS" state, never able to be accessed or re-uploaded.
    - Solution: added Cloudflare Workflow for AI content rating. Workflows can run up to 5 minutes, auto retry step(s) if configured.
- 2nd Problem: filmstrip and other assets from R2 respond intermittently with http 500 errors, because the D1/prisma lookup for the asset's content rating (we only want to serve safe assets so we need to ensure the content_rating for each asset is "SAFE") overwhelms the database pool connection. 
    - Solution: make a 2nd and always public R2 bucket for "SAFE" content that will have a custom domain endpoint. This public endpoint will serve assets instead of D1/the Worker. I've already made and configured the custom domains for the public R2 buckets (staging and prod) using the production UI dashboard. 
    - Solution should work because essentially "if someone has a testId, they could already read all the files. The public bucket doesn't open any new attack surface — it just serves the same already-public data without a Worker in the middle" (argument from AI)  
    
### New upload flows for different environments: 
**Dev, AI off (default)**
- Upload → files written to local private R2, no workflow, test accessible immediately (AI disabled so no rating gate)
- Serving → All assets served through /api/tests/* → local Worker → local private R2

**Dev, AI on (.dev.vars override)**
- Upload → files written to local private R2, workflow runs locally, AI rates it, if SAFE copies to local public R2
- Serving → PUBLIC_ASSETS_URL absent → hasPublicBucket() false → resolveAssetBase returns '' → all assets still served through /api/tests/* → local Worker → local private R2 (public R2 is ignored for serving)

**Staging**
- Upload → files written to remote private R2, workflow triggers on Cloudflare, if SAFE copies to remote public R2
- Serving → For SAFE tests: resolveAssetBase finds config.json in public R2 → screenshot/filmstrip/video URLs built as https://assets-staging.telescopetest.io/... in page HTML → browser loads them directly from the edge, no Worker
- For pre-workflow tests (not yet in public R2): falls back to /api/tests/* → Worker → private R2

**Production**
- Same as staging, https://assets.telescopetest.io instead

### Follow-up 
We'll need some kind of one-time script or process to move all the current R2 files that are safe into the public bucket. Right now, the code has a fallback to use D1/the Worker if the asset isn't in the public bucket. This is so the PR doesn't break any existing results/functionality when merged. 

But I think ideally we'd do this process: (1) merge this PR (with fallback code) -> (2) run one-time script to move files -> (3) clean up fallback code


packages/telescope-web/wrangler.jsonc
```diff
@@ -1,106 +1,140 @@
...
-				"binding": "AI",
-			},
-		},
-	},
-}
\ No newline at end of file
+  "$schema": "node_modules/wrangler/config-schema.json",
+  "name": "telescopetest-io",
+  "account_id": "4b1c95badf17540db82cd8b431ae3eb0",
+  "main": "./src/worker.ts",
```
Judyzc, line 5: `./src/worker.ts` is required to export AiContentRatingWorkflow as a named export alongside Astro's request handler — the standard @astrojs/cloudflare/entrypoints/server entrypoint doesn't support additional exports (https://docs.astro.build/en/guides/integrations-guide/cloudflare/#changed-custom-entrypoint-api).

packages/telescope-web/src/pages/api/tests/[testId]/downloadZIP.ts
```diff
@@ -25,12 +25,18 @@ export const GET: APIRoute = async (context: APIContext) => {
   const bucket = env.RESULTS_BUCKET;
   const prefix = `${testId}/`;
   try {
-    // Use R2 list() function that matches the prefix: https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2listoptions
-    const listed = await bucket.list({ prefix });
-    if (!listed.objects || listed.objects.length === 0) {
+    // Paginate R2 list() — returns at most 1000 objects per call
+    // https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2listoptions
+    let listed = await bucket.list({ prefix });
```
Comment (Judyzc) [inline packages/telescope-web/src/pages/api/tests/[testId]/downloadZIP.ts:30]: Now get R2 assets from public bucket if safe (bypass Worker completely).

packages/telescope-web/src/pages/api/tests/[testId]/[...filename].ts
```diff
@@ -29,23 +33,29 @@ export const GET: APIRoute = async (context: APIContext) => {
...
-  if (aiEnabled) {
-    const rating = await checkTestRating(context, testId);
-    if (rating !== ContentRating.SAFE) {
-      return new Response('Test file not available', { status: 404 });
+  const key = `${testId}/${normalizedFilename}`;
+
+  // Try public bucket first — SAFE files are copied here by the workflow, no rating check needed.
+  // Serve directly (no redirect) since fetch() calls can't follow cross-origin redirects (CORS).
+  const publicFile = hasPublicBucket()
+    ? await env.PUBLIC_RESULTS_BUCKET!.get(key).catch(() => null)
```
Comment (Judyzc) [inline packages/telescope-web/src/pages/api/tests/[testId]/[...filename].ts]: Now get R2 assets from public bucket if safe (bypass Worker completely).

joesanchezjr (approved):

Issue #243, opened 2026-04-16
Performance information not captured when multiple requests share the same URL


sufian-cf:
## Describe the bug

When telescope adds performance data to the HAR, we are matching based on URL. This doesn't work when there are multiple requests to the same URL.

## Telescope info

Please include the following information about your environment:

**Telescope version**: 690199fdb2aa4ce7af3065760ea28560b68f1775
**Node version**: v22.19.0
**Operating system version**: Mac OSX (`Darwin KW4LX2YD22 25.4.0 Darwin Kernel Version 25.4.0`)

## Steps to reproduce

An automated test can trigger this failure:

```diff
diff --git a/__tests__/duplicateRequests.test.ts b/__tests__/duplicateRequests.test.ts
new file mode 100644
index 0000000..b80c673
--- /dev/null
+++ b/__tests__/duplicateRequests.test.ts
@@ -0,0 +1,95 @@
+import type { AddressInfo } from 'node:net';
+
+import { createServer, type Server } from 'node:http';
+import { readFile } from 'node:fs/promises';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { beforeAll, afterAll, describe, expect, test } from 'vitest';
+
+import type { BrowserName, HarEntry } from '../src/types.js';
+import type { SuccessfulTestResult } from '../src/index.js';
+
+import { retrieveHAR } from './helpers.js';
+
+import { launchTest } from '../src/index.js';
+import { BrowserConfig } from '../src/browsers.js';
+
+const browsers: BrowserName[] = BrowserConfig.getBrowsers();
+
+let server: Server;
+let baseUrl: string;
+
+beforeAll(async () => {
+  const allowedFiles = ['index.html', 'style.css'];
+
+  const fixturesDir = join(
+    dirname(dirname(fileURLToPath(import.meta.url))),
+    'tests',
+    'duplicate-requests',
+  );
+  server = createServer(async (req, res) => {
+    const fileName = allowedFiles.find(
+      f => req.url === '/' + f || (f === 'index.html' && req.url === '/'),
+    );
+    if (!fileName) {
+      res.writeHead(404);
+      res.end('Not found');
+      return;
+    }
+    const filePath = join(fixturesDir, fileName);
+    try {
+      const data = await readFile(filePath);
+      const mimeTypes: Record<string, string> = {
+        html: 'text/html',
+        css: 'text/css',
+      };
+      const ext = filePath.split('.').pop() ?? '';
[...]

PR #244, opened 2026-04-16
Fix issue #243, where multiple requests not handled correctly

sufian-cf:
This change fixes issue #243 by adding a new `x-telescope-id` header, containing a unique uuid, to _each_ request made by telescope.

The performance data we're capturing needs to be associated with the HAR, but there isn't a good way to do that without having some sort of unique identifier. We previously were using URL, but that doesn't work if multiple requests are made to the same URL in a telescope test run.

So by adding a header, we can cross-reference the data correctly, and have an added benefit for debugging server-side, so users can identify which requests went sideways via this header (it is present in the generated HAR, and will be sent to servers as `x-telescope-id`).

Note: this _also_ adds package-lock.json changes which should have been added in commit 690199fdb2aa4ce7af3065760ea28560b68f1775 (msw moved to devDependencies).


src/testRunner.ts
```diff
@@ -306,10 +308,38 @@ class TestRunner {
...
+      // Flattening to Record<string, string> can silently drop duplicate
+      // header names (e.g. multiple Set-Cookie entries). This is acceptable
+      // here because route.fallback() requires this shape, and duplicate
+      // request headers are rare in practice.
+      const headers: Record<string, string> = {};
+      for (const { name, value } of original) {
+        headers[name] = value;
+      }
+      headers[TELESCOPE_ID_HEADER] = id;
+      await route.fallback({ headers });
```
sufian-cf: This was a bit annoying. It is technically possible (but unlikely) for requests to have multiple request headers with the same type. And it can't be worked around because of how `.fallback()` works.

I doubt this'll be a real issue we run into, but left this comment for posterity.

src/types.ts
```diff
@@ -428,9 +428,8 @@ export interface RequestTiming {
  * Request data collected during test
  */
 export interface RequestData {
-  url: string;
   timing: RequestTiming;
-  rawTimings?: boolean;
```
sufian-cf, line 438: This `rawTimings` field was always `false` and never used, so I've removed it.

src/types.ts
```diff
@@ -428,9 +428,8 @@ export interface RequestTiming {
  * Request data collected during test
  */
 export interface RequestData {
-  url: string;
```
sufian-cf, line 436: Similarly, we don't need this anymore, since we can match on `telescopeId` exactly

src/testRunner.ts
```diff
@@ -306,10 +308,38 @@ class TestRunner {
...
+    });
+
     page.on('requestfinished', data => {
+      const telescopeId = data.headers()[TELESCOPE_ID_HEADER];
```
sufian-cf, line 332: When we measure performance information, we can now associate the timing information with the exact request id.

mjkozicki (dismissed):

src/testRunner.ts
mjkozicki: Is there any way to log or call this out?

sufian-cf: Ah yeah that's a good idea, I'll add a warning log if that happens

sufian-cf: Actually better is RFC 9110 allows for combining duplicate headers with a comma. Don't need to worry about the Set-Cookie header since that's a response header and we're doing this for request headers.

src/testRunner.ts
```diff
@@ -317,13 +317,16 @@ class TestRunner {
...
-      // header names (e.g. multiple Set-Cookie entries). This is acceptable
-      // here because route.fallback() requires this shape, and duplicate
-      // request headers are rare in practice.
       const headers: Record<string, string> = {};
       for (const { name, value } of original) {
-        headers[name] = value;
+        if (headers[name]) {
+          // Note: We need to flatten an array of headers to a Record<string, string>
+          // If there are duplicates, we can join them with a comma
+          // This is valid per RFC 9110, section 5.2 + 5.3
```
sufian-cf, line 321: https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi

__tests__/duplicateRequests.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+beforeAll(async () => {
+  const allowedFiles = ['index.html', 'style.css'];
+
+  const fixturesDir = join(
+    dirname(dirname(fileURLToPath(import.meta.url))),
+    'tests',
+    'duplicate-requests',
+  );
+  server = createServer(async (req, res) => {
```
sergeychernyshev: Probably just future wish-list, maybe we should extract the generic server for tests rather than duplicating it across tests (looks like we already use it twice).

__tests__/duplicateRequests.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      try {
+        const har = retrieveHAR(testId);
+        expect(har).not.toBeNull();
+
+        const styleCssEntries = har!.log.entries.filter((entry: HarEntry) =>
+          entry.request.url.endsWith('/style.css'),
+        );
+
+        // The page loads style.css via <link> and fetches it twice via JS,
```
sergeychernyshev: This seems to be quite uncertain about the results we expect here.
Should the exact amount be recorded (even if some served from cache)?

Is there a way to be certain here?

__tests__/hostOverride.test.ts
```diff
@@ -0,0 +1,119 @@
...
+});
+
+afterAll(async () => {
+  await new Promise<void>(resolve => server.close(() => resolve()));
+});
+
+describe.each(browsers)(
+  'Host override with HAR timing correlation - %s',
+  (browser: BrowserName) => {
+    test('HAR entries have timing data when overrideHost is used', async () => {
```
sergeychernyshev: If this is specific to duplicate request issue (which is true, as I understand it), I'd move these test cases to `duplicateRequests.test.ts`

If not, then we should probably not include it in this PR.

src/testRunner.ts
sergeychernyshev: I'd love to not modify existing headers if all we need is to just add a new one.

It looks like implementation collapses them too (method is literally called [`_fromHeadersObjectLossy()`](https://github.com/microsoft/playwright/blob/1112f6e3752b52ec9cdc2efa86370f978b617e78/packages/playwright-core/src/client/network.ts#L915-L920)

sufian-cf: Ah yeah, this a hard limitation baked into the Playwright API for `.fallback()` -- the only way to adjust headers is to provide a `Record<string, string>` -- so if there were multiple headers in the array, they need to be combined into a single header. (Or we drop one, which is what this did prior to the most recent commit in this PR -- I wasn't a fan of that, so figured we combine them per the http spec rfc)

__tests__/duplicateRequests.test.ts
sufian-cf: Yeah, there's a lot that could be cleaned up in our tests -- I figure we could do that separately

__tests__/duplicateRequests.test.ts
sufian-cf: Fair enough, was being cautious due to different browsers possibly behaving differently, just updated this to be precise.

src/testRunner.ts
sergeychernyshev: I tried to do more ai-assisted research and there seems to be no way for a better per-request solution - the only alternative is `setExtraHTTPHeaders` which is global for all requests.

There are multiple stages and you don't really get all headers to begin with and browsers only have "setHeaders" APIs, but one would've thought at least I can fully manipulate the original source and provide it back.

Anyway, we could probably just document the impact of collapsing the headers into one.

sergeychernyshev (dismissed):
I feel like this is good-enough scenario.

Alternatively, we can make this conditional somehow... e.g. allow disabling this "higher precision" request reconciliation at the expense of keeping headers intact.

src/testRunner.ts
sufian-cf: I've confirmed we cannot avoid collapsing headers in playwright and also confirmed that this does not change request headers unless there are duplicates. (the above comment was a mistake, there is no case normalization) -- I've updated that comment.

sergeychernyshev (approved):
LTGM

PR #295, opened 2026-06-10
config command line option

### LINKED ISSUE #62: Add an option to provide a config file.
Add an option to specify a configuration file containing the values used to pass to the playwright engine. The config.json file that is output to the results directory should be reusable to launch a test with the same parameters.

CLI parameters override the values in the config.

### Proposed format
`--config <file_path>. Provide a configuration file to initiate the test.`

The config.json file can include a `config` property that stores all the specified values and no defaults.
If the `config` property does not exist, fallback creating the test using all values in the config.json.

### Examples
```bash
npx . -u https://example.com
```
produces
```json
"runParameters": {
  "url": "https://example.com"
}
```
---
```bash
npx . -u https://example.com -b "firefox" -h {"header1":"one","header2":"two"}
```
produces
```json
"runParameters": {
  "url": "https://example.com",
  "browser": "firefox",
  "headers": {"header1":"one","header2":"two"}
}
```

ozcoder:
Added the command line option `--config CONFIG_FILENAME`. Command line options will override any options set in the configuration file.

I've moved some things around, in particular _defaults_ are handled in `executeTest`. The URL to test can be in the configuration file or `-u URL` or as an argument to the command line.

Closes https://github.com/cloudflare/telescope/issues/62


packages/telescope/src/index.ts
```diff
@@ -140,13 +140,36 @@ function getRunner(
...
+  // (e.g. `telescope example.com` -> `https://example.com`, or
+  // `telescope localhost:3000` -> `http://localhost:3000`; see
+  // normalizeUrlScheme for the rules). Reject non-http(s) URLs up front so
+  // the user gets a clearer error than the deeper validation in executeTest
+  // would produce. CLI-only convenience; programmatic callers (launchTest,
+  // Telescope) must provide a well-formed http(s) URL.
+  try {
+    options.url = normalizeUrlScheme(resolvedUrl);
+  } catch (err) {
+    throw err;
```
nicjansma: `npm run lint` is throwing an error with this, the try/catch isn't necessary if it just throws.

packages/telescope/src/index.ts
```diff
@@ -460,26 +489,26 @@ export default function browserAgent(): void {
...
+  // // normalizeUrlScheme for the rules). Reject non-http(s) URLs up front so
+  // // the user gets a clearer error than the deeper validation in executeTest
+  // // would produce. CLI-only convenience; programmatic callers (launchTest,
+  // // Telescope) must provide a well-formed http(s) URL.
+  // try {
+  //   cliOptions.url = normalizeUrlScheme(resolvedUrl);
+  // } catch (err) {
+  //   console.error(`error: ${(err as Error).message}`);
+  //   process.exit(1);
+  // }
```
nicjansma: Was all this just moved higher in the function?  We can remove then, instead of commenting out.

packages/telescope/src/config.ts
```diff
@@ -165,11 +165,120 @@ export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
...
+  const engine = browserConfig.engine; // Required
+
+  if (engine === 'firefox') {
+    cfg.browser = 'firefox';
+    if (browserConfig.firefoxUserPrefs) {
+      cfg.firefoxPrefs = browserConfig.firefoxUserPrefs;
+    }
+  } else if (engine === 'webkit') {
+    cfg.browser = 'safari';
+  } else { // chromium
```
nicjansma, line 248: Can you run `npm run prettier` on the project?  I think some of these minor formatting consistency things will be auto-fixed.

packages/telescope/src/config.ts
```diff
@@ -165,11 +165,120 @@ export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
...
+      throw new Error(`Can not read file ${configFileName}`);
+    }
+  });
+
+  stat(configFileName, (err, cfgStat) => {
+    if (err) {
+      throw err;
+    }
+
+    if (cfgStat.size > 4096) { // Arbitrary size
```
nicjansma: Any particular reason for blocking at > 4096 bytes?

I could see things like headers potentially grow above this.

packages/telescope/__tests__/emulation.test.ts
```diff
@@ -275,19 +275,4 @@ describe('Device default browser resolution', () => {
...
-      });
-      expect(config_options.browser).toBe('chrome');
-
-      const config = new BrowserConfig().getBrowserConfig(
-        config_options.browser!,
-        config_options,
-      );
-      expect(config.engine).toBe('chromium');
-    });
-  });
```
nicjansma, line 292: Will the default not be Chromium anymore?

packages/telescope/src/config.ts
```diff
@@ -165,11 +165,120 @@ export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
...
+ */
+
+export function getBaseConfig(configFileName: string): LaunchOptions {
+  let baseConfig: ConfigFileType = { };
+
+  access(configFileName, constants.R_OK, (err) => {
+    if (err) {
+      throw new Error(`Can not read file ${configFileName}`);
+    }
+  });
```
nicjansma: I think `node:fs` `access()` and `stat()` will be async here, so will start those functions but quickly continue to the try block below with the sync `readFileSync()`.

In other words, the error callbacks from checking access/stat won't fire before the `readFileSync` fires, so the errors we see will be from the throw below and not the intentional throws here, nor will `cfgStat.size` trigger.

packages/telescope/src/index.ts
```diff
@@ -436,9 +455,19 @@ export default function browserAgent(): void {
...
+      new Option(
+        '--config <string>',
+        'Use a configuration file for the options',
+      )
+    )
+    // .action((opts) => {
+    //   if (!opts.url && !opts.config) {
+    //     program.error(`Error: - required option '-u, --url <url>' not specified`);
+    //   }
+    // })
```
nicjansma: Remove if unused

packages/telescope/src/index.ts
```diff
@@ -140,13 +140,36 @@ function getRunner(
...
   options: LaunchOptions,
+  baseConfig: LaunchOptions
 ): Promise<SuccessfulTestResult> {
+  const positionalUrl = program.args[0];
+  const resolvedUrl = positionalUrl ?? (options.url || baseConfig.url);
+  if (!resolvedUrl) {
+    console.error(
+      "error: missing required URL argument. Use 'telescope --help' for usage.",
+    );
+    process.exit(1);
```
nicjansma: The existing `executeTest()` contract is to throw on error, and may break API access to this function if `process.exit(1)` is called.  `process.exit()` should only be in `browserAgent()`.

packages/telescope/src/index.ts
```diff
@@ -140,13 +140,36 @@ function getRunner(
...
+ *        baseConfig - Test options from a config file
  * @returns Test result with ID and results path
  * @throws If the test fails
  * @private
  */
 async function executeTest(
   options: LaunchOptions,
+  baseConfig: LaunchOptions
 ): Promise<SuccessfulTestResult> {
+  const positionalUrl = program.args[0];
```
nicjansma: `program.` querying should only be used in `browserAgent()`.

packages/telescope/src/config.ts
```diff
@@ -165,11 +165,120 @@ export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
...
+  let cfg: LaunchOptions = { url: baseConfig.url || '' }; // Always required
+
+  if (baseConfig.options) {
+    type CommonCLIKeys = Extract<keyof ConfigCLIOptionsType, keyof LaunchOptions>;
+
+    for (const key of Object.keys(ConfigCLIOptionsSchema.shape) as CommonCLIKeys[]) {
+      const value = baseConfig.options[key];
+      if (value !== undefined) {
+        // Safe because of the Extract above
+        cfg[key] = value as any; // eslint-disable-line
```
nicjansma: We should be enforcing `@typescript-eslint/no-explicit-any`, can this be done another way?

packages/telescope/src/index.ts
```diff
@@ -307,7 +342,6 @@ export default function browserAgent(): void {
     )
     .addOption(
       new Option('--delayUsing <string>', 'Method to use to delay responses')
-        .default(DEFAULT_OPTIONS.delayUsing)
```
nicjansma, line 310: A big downside to removing the defaults here is that the CLI `--help` no longer shows them.

Can precedence be handled elsewhere?

packages/telescope/src/config.ts
```diff
@@ -40,29 +45,24 @@ const ENGINE_TO_BROWSER_NAME: Record<PlaywrightEngine, BrowserName> = {
  * @throws If `options.url` is not set (programmer error: caller broke contract)
  */
 export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
-  if (!options.url) {
-    throw new Error('normalizeCLIConfig: url is required');
-  }
```
nicjansma, line 45: Any reason URL checking here needs to be fully removed?

packages/telescope/src/index.ts
```diff
@@ -436,9 +455,19 @@ export default function browserAgent(): void {
         '--device <string>',
         'Device to use for device emulation (viewport size, DPR, touch events). Also sets the default browser engine unless overridden with -b. Devices are based on the Playwright device list (see https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json)',
       ),
-    );
-
-  program.parse(process.argv);
+    )
+    .addOption(
+      new Option(
+        '--config <string>',
```
nicjansma, line 453: Can you update `README.md` for the new `--config` option?

packages/telescope/src/schemas.ts
```diff
@@ -38,3 +38,78 @@ export const StringArraySchema = z.array(z.string());
...
+});
+
+/**
+ * Schema for an input config file
+ **/
+
+export const ConfigFileSchema = z.object({
+  url: z.string().refine((uri) => { // Zod v3
+    if (!(/^https?:\/\//).test(uri)) {
+      uri = `https://${uri}`; // Append missing protocol so new URL can work
```
nicjansma, line 104: This modifies the `uri` for validation but doesn't actually apply the change right?

nicjansma (changes requested):
Thanks for the PR @ozcoder!  Would love to have this option.

packages/telescope/src/config.ts
```diff
@@ -165,11 +165,120 @@ export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
...
+    const cfgObject = JSON.parse(cfgData);
+    baseConfig = ConfigFileSchema.parse(cfgObject);
+  } catch (err) {
+    console.error(err);
+    throw new Error(`Problem parsing ${configFileName}`);
+  }
+
+  let cfg: LaunchOptions = { url: baseConfig.url || '' }; // Always required
+
+  if (baseConfig.options) {
```
nicjansma, line 206: It wasn't immediately obvious to me that the config JSON file should have a top-level `options: {}` key that all CLI options go into.

After looking at the original issue, I see that this comes from the `results\xxx\config.json` having a top-level `options: {}`.

Maybe this can be just documented clearly, with an example, in `README.md`?

packages/telescope/src/config.ts
ozcoder: Good point.
Arbitrary size, maybe 32K would be better. I was hoping for some guidance.

packages/telescope/src/index.ts
ozcoder: Oops, forgot to clean up before checking in. Sorry.

packages/telescope/src/index.ts
ozcoder: Oops, sorry. Forgot to tidy up before committing.

packages/telescope/src/schemas.ts
ozcoder: Because we are using Zod version 3.* we don't have a built in URL checking when it doesn't have the protocol. So, I wrote a simple one. The `refine` method doesn't alter the `url` variable, it just checks and needs to return a _falsy_ value to reject it as a _schema fail_. The message in a object counts as _falsy_. 
Code later on will really add the `https://` if needed that somebody else wrote.

packages/telescope/src/config.ts
ozcoder: You are correct. I changed this a few times, but I should be using the Sync versions.

packages/telescope/src/config.ts
ozcoder: Can't _truely_ get away from _any_, but I have rewritten this to not need the `eslint-disable-line`.

packages/telescope/src/config.ts
ozcoder: The URL can be given in the config file, and/or command line option `-u` and/or command line option `--url` and/or a positional argument, so line 43 would fail if it is **only** in the config file. 
Command line options override what is in the config file though.
Zod will check if the url given _looks like_ it _could_ be a valid URL.

Note: Some people have single word hostnames on their local network, that are valid. 
Eg. https://books/ would be their local audiobook server.

packages/telescope/src/schemas.ts
ozcoder: I've rewritten this a little bit better.

packages/telescope/src/index.ts
ozcoder: good point

packages/telescope/src/index.ts
ozcoder: I found a way to display the defaults in the help messaging.

packages/telescope/__tests__/emulation.test.ts
ozcoder: `normalizeCLIConfig` no longer sets all the defaults, so I have created `resolveOptions` that merges the config file values on top of the defaults, and then merges the command line/programmatic on top of that, to come up with the config. The default is still `chrome`. I have fixed the tests to also call `resolveOptions` to get the final configuration, which will include some defaults.

packages/telescope/src/index.ts
```diff
@@ -228,7 +260,8 @@ async function executeTest(
  */
 export async function launchTest(options: LaunchOptions): Promise<TestResult> {
   try {
-    return await executeTest(options);
+    const baseConfig = options.config ? getBaseConfig(options.config) : { url: '' };
+    return await executeTest(options, baseConfig);
   } catch (error) {
```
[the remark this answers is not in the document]
ozcoder: I need help creating a test. I'm not sure where the test config should live.

packages/telescope/src/config.ts
nicjansma: 32K seems more reasonable.

What kind of situations do you think the size check will help with?  People putting incorrect data in there for some reason?  Unsure if we need the gate at all.

packages/telescope/src/index.ts
```diff
@@ -342,63 +358,60 @@ export default function browserAgent(): void {
...
+      ).default(null, `${DEFAULT_OPTIONS.frameRate}`
+      ).argParser(v => parseNumeric(PositiveIntSchema, v, '--frameRate')),
     )
     .addOption(
-      new Option('--disableJS', 'Disable JavaScript').default(
-        DEFAULT_OPTIONS.disableJS,
-      ),
+      new Option(
+        '--disableJS',
+        `Disable JavaScript (default: ${DEFAULT_OPTIONS.disableJS})`)
```
nicjansma, line 381: Is there a reason some of the options use `.default(xyz)` and others describe the default in the description `... (default: xyz)`?

packages/telescope/src/index.ts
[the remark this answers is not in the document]
nicjansma: I think a clone of `params.test.ts` that focuses on the `--config` options would be nice.  I haven't written tests for this project yet either!  But if you have access to a coding agent, it could probably start with some boilerplate.

nicjansma:
Thanks for the updates @ozcoder!  Lots of the changes look good.

Only a few outstanding things left:

* Updates to README to describe `--config`
* `npm run prettier` and `npm run lint:fix`
* tests

Thanks!

packages/telescope/src/index.ts
ozcoder: Unfortunately, the `.default` method does not work with booleans when we just want to display what the _default_ is for that option, so I have to use a template literal.

PR #305, opened 2026-07-02
Include LICENSE in both telescope and waterfall packages

nicjansma:
* Include LICENSE in both telescope and waterfall packages, so they're in the NPM published packages
* Bump Copyright to 2026


sergeychernyshev (commented):
Does it get packaged when npm package is published?

nicjansma:
> Does it get packaged when npm package is published?

The root-of-the-repo `LICENSE` is not included with `npm pack` as it won't pull in any files outside of that packages' root directory, so we need to replicate it for each package dir to be automagically pulled in.

ryantownsend:
> Does it get packaged when npm package is published?

The `package.json` include the `"license"` key which should cover this, right?

https://github.com/cloudflare/telescope/blob/af6a11cf52f876992714d2e7183236146933bef1/package.json#L31
https://github.com/cloudflare/telescope/blob/af6a11cf52f876992714d2e7183236146933bef1/packages/waterfall/package.json#L66
https://github.com/cloudflare/telescope/blob/af6a11cf52f876992714d2e7183236146933bef1/packages/telescope/package.json#L52

The exception is `telescope-web` is private: https://github.com/cloudflare/telescope/blob/af6a11cf52f876992714d2e7183236146933bef1/packages/telescope-web/package.json#L3

nicjansma:
It does show the license on [npmjs.com](https://www.npmjs.com/package/@cloudflare/telescope)

image

But it's not a file living within the repo that contains the Copyright dates on it, which the `LICENSE` file has.

sergeychernyshev (approved):

PR #317, opened 2026-07-21
Extract external CSS stylesheets from HAR for Baseline feature detection

smsnedden:
First slice of the `--baseline` CSS analysis feature. This PR adds extraction of **external stylesheets** from a HAR; inline `<style>` extraction, feature detection (`css-tree`), and Baseline lookup (`web-features`) follow in later PRs.

The extractor is intentionally **not wired into the CLI or `launchTest`** yet — it has no production caller until the detection PR consumes it, so it's exercised entirely by its unit tests. Keeping the slices separate keeps each PR small and isolates new dependencies to the PR that introduces them.

### `extractCSSFromHar(harData): CSSSource[]` (`src/baselineCssExtract.ts`)

Collects external CSS a page shipped, from a parsed HAR:

- Selects responses whose `mimeType` starts with `text/css` (case-insensitive, so `Text/CSS; charset=utf-8` matches), using the response body as the CSS source.
- Decodes base64 response bodies to UTF-8 (when HAR `content.encoding === 'base64'`).
- Returns sources in HAR order; entries with no response body are skipped.

Inline `<style>` blocks are out of scope here — they require parsing the HTML document, added in a follow-up PR with a dedicated parser.

### Types (`src/types.ts`)

- Add `CSSSource` (`{ css, file }`).
- Extend `HarEntry.response.content` with optional `text?` and `encoding?` fields (response bodies weren't previously modeled).

## Tests

Adds `packages/telescope/__tests__/baselineCssExtract.test.ts` — 7 pure unit tests (no browser) built from in-memory HAR fixtures:

- External `text/css` extracted with the URL as `file`
- `text/css; charset=utf-8` still matches
- base64-encoded CSS bodies are decoded
- Multiple stylesheets preserve HAR order
- Edge cases: empty HAR, entries with no body, and non-CSS resources (including HTML) ignored

### DIFF
--- a/packages/telescope/__tests__/baselineCssExtract.test.ts
+++ b/packages/telescope/__tests__/baselineCssExtract.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from 'vitest';
+
+import { extractCSSFromHar } from '../src/baselineCssExtract.js';
+import type { HarData, HarEntry } from '../src/types.js';
+
+// Build a minimal HAR entry in-memory (only the fields the extractor reads).
+function makeEntry(
+  url: string,
+  mimeType: string,
+  text: string | undefined,
+  encoding?: string,
+): HarEntry {
+  return {
+    request: { url, method: 'GET', headers: [] },
+    response: {
+      status: 200,
+      content: { size: text?.length ?? 0, mimeType, text, encoding },
+    },
+    time: 0,
+    startedDateTime: '1970-01-01T00:00:00.000Z',
+  };
+}
+
+function makeHar(entries: HarEntry[]): HarData {
+  return {
+    log: { pages: [], entries, browser: { name: 'test', version: '0' } },
+  };
+}
+
+const toBase64 = (value: string) =>
+  Buffer.from(value, 'utf8').toString('base64');
+
+describe('extractCSSFromHar — external stylesheets', () => {
+  it('extracts a text/css response as one source with the URL as file', () => {
+    const har = makeHar([
+      makeEntry('https://x.test/a.css', 'text/css', '.a { color: red; }'),
+    ]);
+
+    expect(extractCSSFromHar(har)).toEqual([
+      { css: '.a { color: red; }', file: 'https://x.test/a.css' },
+    ]);
+  });
+
+  it('matches text/css even with a charset parameter', () => {
+    const har = makeHar([
+      makeEntry('https://x.test/a.css', 'text/css; charset=utf-8', '.a {}'),
+    ]);
+
+    expect(extractCSSFromHar(har)).toHaveLength(1);
+  });
+
+  it('decodes base64-encoded CSS bodies', () => {
+    const css = '.a { color: blue; }';
+    const har = makeHar([
+      makeEntry('https://x.test/a.css', 'text/css', toBase64(css), 'base64'),
+    ]);
+
+    expect(extractCSSFromHar(har)[0].css).toBe(css);
+  });
+
+  it('preserves entry order across multiple stylesheets', () => {
+    const har = makeHar([
+      makeEntry('https://x.test/1.css', 'text/css', '.one {}'),
+      makeEntry('https://x.test/2.css', 'text/css', '.two {}'),
+    ]);
+
+    expect(extractCSSFromHar(har).map(source => source.file)).toEqual([
+      'https://x.test/1.css',
+      'https://x.test/2.css',
+    ]);
+  });
+});
+
+describe('extractCSSFromHar — ignored and edge cases', () => {
+  it('returns an empty array when there are no entries', () => {
+    expect(extractCSSFromHar(makeHar([]))).toEqual([]);
+  });
+
+  it('ignores entries with no response body text', () => {
+    const har = makeHar([
+      makeEntry('https://x.test/a.css', 'text/css', undefined),
+    ]);
+
+    expect(extractCSSFromHar(har)).toEqual([]);
+  });
+
+  it('ignores non-CSS resources', () => {
+    const har = makeHar([
+      makeEntry('https://x.test/img.png', 'image/png', 'notcss'),
+      makeEntry(
+        'https://x.test/app.js',
+        'application/javascript',
+        'const a = 1;',
+      ),
+      makeEntry('https://x.test/', 'text/html', '<html></html>'),
+    ]);
+
+    expect(extractCSSFromHar(har)).toEqual([]);
+  });
+});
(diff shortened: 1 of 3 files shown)


packages/telescope/src/types.ts
```diff
@@ -571,6 +571,8 @@ export interface HarEntry {
     content: {
       size: number;
       mimeType: string;
+      text?: string;
```
sergeychernyshev, line 585: Why are these optional? do we not collect all bodies? what controls that?

packages/telescope/src/types.ts
```diff
@@ -571,6 +571,8 @@ export interface HarEntry {
...
+      encoding?: string;
```
sergeychernyshev: What are the possible values here? should this be an enum instead?

packages/telescope/src/baselineCssExtract.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Decode a HAR response body, handling base64-encoded content.
+ *
+ * @param text - The raw `content.text` value from a HAR entry.
+ * @param encoding - The `content.encoding` value, if present.
+ * @returns The decoded body as UTF-8 text.
+ */
+function decodeContent(text: string, encoding?: string): string {
+  if (encoding === 'base64') {
```
sergeychernyshev: Probably better to use TypeScript enum if we only expect `base64` as a value.

sergeychernyshev (changes requested):

sergeychernyshev:
Looks good. Just minor changes in typing, also need to understand in general if HAR will always contain bodies or not and what controls it.

packages/telescope/src/types.ts
smsnedden: They're optional because response bodies aren't always guaranteed. Redirects, 204s, or failed requests might not have `text`, and `encoding` only appears when `text` is present and base64-encoded. Telescope does collect response bodies by default via Playwright's `recordHar`, but the type still needs to accommodate entries that legitimately lack them.

packages/telescope/src/types.ts
smsnedden: Looking at the [HAR 1.2 spec](http://www.softwareishard.com/blog/har-12-spec/), `encoding` and `text` are defined as `[string, optional]`. Since we're consuming external HARs, keeping both as `string` matches the spec and avoids being too restrictive.

sergeychernyshev: To meet it looks like optional "base64" value. A good candidate for a const enum, but not critical.

At least that's how you use it in your code so let's define it as such in types - if we start using it differently, we could.

Remember that TypeScript types are only for type definitions in your code - actual executable code knows nothing about types.

smsnedden: Sounds good, I'll make that change

sergeychernyshev (dismissed):
Looks good - up to you if you want to do a bit more strict type checks here.

smsnedden:
> Looks good - up to you if you want to do a bit more strict type checks here.

Sounds good, made a small change to line 39 for a bit more strict of a type check.

packages/telescope/src/baselineCssExtract.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Decode a HAR response body, handling base64-encoded content.
+ *
+ * @param text - The raw `content.text` value from a HAR entry.
+ * @param encoding - The `content.encoding` value, if present.
+ * @returns The decoded body as UTF-8 text.
+ */
+function decodeContent(text: string, encoding?: 'base64'): string {
```
sergeychernyshev: Let's move this as a constant to `types.ts` and create a type and constant, something like

```
export const BASE64 = 'base64';
export type HARContentEncoding = typeof BASE64 | undefined;
```

```suggestion
function decodeContent(text: string, encoding: HARContentEncoding): string {
```

packages/telescope/src/baselineCssExtract.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (encoding === 'base64') {
+    return Buffer.from(text, 'base64').toString('utf8');
```
sergeychernyshev:
```suggestion
    return Buffer.from(text, BASE64).toString('utf8');
```

packages/telescope/src/baselineCssExtract.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Decode a HAR response body, handling base64-encoded content.
+ *
+ * @param text - The raw `content.text` value from a HAR entry.
+ * @param encoding - The `content.encoding` value, if present.
+ * @returns The decoded body as UTF-8 text.
+ */
+function decodeContent(text: string, encoding?: 'base64'): string {
+  if (encoding === 'base64') {
```
sergeychernyshev:
```suggestion
  if (encoding === BASE64) {
```

packages/telescope/src/types.ts
```diff
@@ -571,6 +571,8 @@ export interface HarEntry {
     content: {
       size: number;
       mimeType: string;
+      text?: string;
+      encoding?: 'base64';
```
sergeychernyshev:
```suggestion
      encoding: HARContentEncoding;
```

packages/telescope/__tests__/baselineCssExtract.test.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function makeHar(entries: HarEntry[]): HarData {
+  return {
+    log: { pages: [], entries, browser: { name: 'test', version: '0' } },
+  };
+}
+
+const toBase64 = (value: string) =>
+  Buffer.from(value, 'utf8').toString('base64');
```
sergeychernyshev:
```suggestion
  Buffer.from(value, 'utf8').toString(BASE64);
```

packages/telescope/__tests__/baselineCssExtract.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      makeEntry('https://x.test/a.css', 'text/css; charset=utf-8', '.a {}'),
+    ]);
+
+    expect(extractCSSFromHar(har)).toHaveLength(1);
+  });
+
+  it('decodes base64-encoded CSS bodies', () => {
+    const css = '.a { color: blue; }';
+    const har = makeHar([
+      makeEntry('https://x.test/a.css', 'text/css', toBase64(css), 'base64'),
```
sergeychernyshev:
```suggestion
      makeEntry('https://x.test/a.css', 'text/css', toBase64(css), BASE64),
```

sergeychernyshev (commented):
Let's use a constant and type instead of repeating the string itself.

The reason for such abstraction is to easily add other options through types, for example, without rewriting everything in the future.

It is a minor optimization, but helpful.

smsnedden:
> Let's use a constant and type instead of repeating the string itself.
> 
> The reason for such abstraction is to easily add other options through types, for example, without rewriting everything in the future.
> 
> It is a minor optimization, but helpful.

Those changes are now made

sergeychernyshev (approved):
Let's go! 🚀

PR #330, opened 2026-07-24
Harvest inline <style> CSS from the live DOM for Baseline feature detection

smsnedden:
Second slice of the `--baseline` CSS analysis feature. This PR adds extraction of **inline `<style>` blocks** by reading them from a page's live DOM; CSS feature detection (`css-tree`) and Baseline lookup (`web-features`) over the collected sources follow in later PRs. (External stylesheets landed in #317.)

Like the external extractor, `harvestInlineStyles` is intentionally **not wired into the CLI or `launchTest`** yet. It is designed to be driven by the **Baseline detection pass** — a live second-pass browser navigation (the same pass that performs JS API detection) — which is productionized in a follow-up PR. Reading inline CSS there keeps all live-DOM detection in one place and one navigation, and runs it once: Baseline analyzes browser-agnostic *source* CSS, so there is no need to repeat it across the browser matrix. Keeping this slice separate keeps each PR small and reviewable.

**Why read the live DOM instead of parsing the HTML?** So that `<style>` blocks a page injects at runtime (e.g. via JavaScript) are captured — static HTML parsing would miss them. The browser is already running in the detection pass, so it is the ground truth for what the page actually applied.

### `harvestInlineStyles(page): Promise<CSSSource[]>` (`src/baselineCssExtract.ts`)

Reads inline CSS from a loaded page's live DOM:

- Collects the text of every `<style>` element in document order, returning one `CSSSource` per non-empty block (the same `{ css, file }` shape as `extractCSSFromHar`).
- Skips empty / whitespace-only blocks.
- The read is strictly **read-only** — a single `page.evaluate` that reads `<style>` text with no DOM mutation and no layout-triggering reads — so it can run inside the shared detection pass without interfering with co-located instrumentation or page state.

**Out of scope** (documented in the JSDoc; may be revisited later): only `<style>` element text is read, so CSS with no `<style>` text — rules added via `CSSStyleSheet.insertRule()` or constructable `adoptedStyleSheets` — is not captured, nor are `<style>` elements inside shadow roots (`querySelectorAll` does not pierce shadow DOM) or `<template>` content.

### Types (`src/types.ts`)

- Document `CSSSource`: `css` is the field that gets parsed; `file` is a provenance label for reporting — a URL for external stylesheets, a page-derived label (e.g. `https://example.com/ (inline style #1)`) for inline blocks. It is a display string, not a URL to parse.

## Tests

Extends `packages/telescope/__tests__/baselineCssExtract.test.ts` with 6 tests parameterized per rendering engine (chromium/firefox/webkit; Firefox-only in CI), driving a real page via `setContent`:

- Single `<style>` block extracted as one source
- Multiple blocks preserved in document order
- **`<style>` injected by JavaScript after load is captured** (the case static parsing misses)
- No `<style>` blocks → empty array
- Empty / whitespace-only blocks skipped
- **Read-only guarantee**: harvesting leaves the document markup unchanged

### DIFF
--- a/packages/telescope/__tests__/baselineCssExtract.test.ts
+++ b/packages/telescope/__tests__/baselineCssExtract.test.ts
@@ -107,22 +107,24 @@ describe('extractCSSFromHar — ignored and edge cases', () => {
 });
 
 // Inline `<style>` blocks are read from the live DOM rather than the HAR, so
-// these tests drive a real page per rendering engine in the selected matrix
-// (CI runs Firefox only; locally all engines run). setContent leaves the page
-// on `about:blank`, which is what `file` reflects.
-const engines = [
-  ...new Set(
-    BrowserConfig.getBrowsers().map(
-      name => BrowserConfig.browserConfigs[name].engine,
-    ),
-  ),
-];
+// these tests drive a real page in each browser from Telescope's own matrix,
+// launched from that browser's Telescope config (engine + channel + headless)
+// so coverage tracks the browsers Telescope actually runs rather than a
+// separate engine list. setContent leaves the page on `about:blank`, which is
+// what `file` reflects.
+const browsers = BrowserConfig.getBrowsers();
 
-describe.each(engines)('harvestInlineStyles — %s', engine => {
+describe.each(browsers)('harvestInlineStyles — %s', browserName => {
+  const config = BrowserConfig.browserConfigs[browserName];
   let browser: Browser;
 
   beforeAll(async () => {
-    browser = await playwright[engine].launch({ headless: true });
+    browser = await playwright[config.engine].launch({
+      headless: config.headless,
+      ...('channel' in config && config.channel
+        ? { channel: config.channel }
+        : {}),
+    });
   }, 120000);
 
   afterAll(async () => {
--- a/packages/telescope/__tests__/baselineCssExtract.test.ts
+++ b/packages/telescope/__tests__/baselineCssExtract.test.ts
@@ -107,11 +107,8 @@ describe('extractCSSFromHar — ignored and edge cases', () => {
 });
 
 // Inline `<style>` blocks are read from the live DOM rather than the HAR, so
-// these tests drive a real page in each browser from Telescope's own matrix,
-// launched from that browser's Telescope config (engine + channel + headless)
-// so coverage tracks the browsers Telescope actually runs rather than a
-// separate engine list. setContent leaves the page on `about:blank`, which is
-// what `file` reflects.
+// these tests drive a real page in each browser from Telescope's matrix.
+// setContent leaves the page on `about:blank`, which is what `file` reflects.
 const browsers = BrowserConfig.getBrowsers();
 
 describe.each(browsers)('harvestInlineStyles — %s', browserName => {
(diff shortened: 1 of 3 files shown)


nicjansma (dismissed):

packages/telescope/src/baselineCssExtract.ts
```diff
@@ -30,6 +32,52 @@ export function extractCSSFromHar(harData: HarData): CSSSource[] {
...
+  const blocks = await page.evaluate(() =>
+    // `textContent` on an element is always a string; `?? ''` only satisfies
+    // its `string | null` type (null is unreachable for these nodes).
+    Array.from(
+      document.querySelectorAll('style'),
+      style => style.textContent ?? '',
+    ),
+  );
+
+  const pageUrl = page.url();
```
sergeychernyshev, line 67: Minor nit - probably better to rewrite using chained `filter()` and `map()` calls.

packages/telescope/__tests__/baselineCssExtract.test.ts
```diff
@@ -98,3 +105,108 @@ describe('extractCSSFromHar — ignored and edge cases', () => {
     expect(extractCSSFromHar(har)).toEqual([]);
   });
 });
+
+// Inline `<style>` blocks are read from the live DOM rather than the HAR, so
+// these tests drive a real page per rendering engine in the selected matrix
+// (CI runs Firefox only; locally all engines run). setContent leaves the page
```
sergeychernyshev: Unless this is critical for specific test, e.g. test only runs in chromium, for example, there is no need to specify what CI does differently here as it might drift as we modify CI which is far enough from this code for people or AI agents to miss the discrepancy.

packages/telescope/__tests__/baselineCssExtract.test.ts
```diff
@@ -98,3 +105,108 @@ describe('extractCSSFromHar — ignored and edge cases', () => {
...
+      name => BrowserConfig.browserConfigs[name].engine,
+    ),
+  ),
+];
+
+describe.each(engines)('harvestInlineStyles — %s', engine => {
+  let browser: Browser;
+
+  beforeAll(async () => {
+    browser = await playwright[engine].launch({ headless: true });
```
sergeychernyshev: I am worried that we limit this to just engines and not deeper browser configurations here which can drift away from telescope browser configurations that would be ran.

Why do we not want to use just fresh instances of the same browsers we use for Telescope tests?

I guess we can live with this testing just CSS extraction in general so we can leave it as is for now...

sergeychernyshev (dismissed):
Major engine vs. browser comment might be worth remembering for further work where we'll be running actual pages and not tests, like here.

The rest are minor nit-picks.

sergeychernyshev:
I see that tests are failing because of package mismatch, you probably want to sync your fork and rebase this branch over it.

packages/telescope/__tests__/baselineCssExtract.test.ts
smsnedden: Good call - I made the switch from launching per-engine to iterating `BrowserConfig.getBrowsers()` and launching each from its Telescope config (engine + channel + headless), so now it tracks the same browsers Telescope actually runs.

I can plan to alter it to the full `getBrowserConfig()` in a later PR (one of the JS detection ones) to make sure it matches up once we have it doing the full second-pass navigation.

sergeychernyshev (approved):
🚀

PR #337, opened 2026-07-24
Add Baseline CLI pipeline skeleton

smsnedden:
Introduces the `--baseline` entry point and the initial Baseline pipeline infrastructure. When enabled, the pipeline runs after the performance navigation has completed, the browser has closed, and the HAR and metrics artifacts have been finalized.

The pipeline is intentionally minimal in this PR: it establishes option plumbing, execution ordering, error isolation, and the artifact-writing convention by producing `baseline/meta.json`. Feature detection, the separate detection navigation, Baseline status lookup, and final status-split reports follow in later PRs.

### `--baseline` option

- Adds a boolean `--baseline` CLI option, defaulting to `false`.
- Threads the option through CLI normalization and the programmatic `launchTest` API.
- Adds the option to `LaunchOptions`, `DefaultOptions`, `CLIOptions`, and the normalized internal `TestConfig`.
- Preserves the normalized `baseline: boolean` guarantee through the browser runners.

### Baseline pipeline (`src/baseline.ts`)

Adds `runBaselinePipeline`, invoked during post-processing after the HAR and metrics have been written.

For this initial slice, it writes:

```text
baseline/meta.json
```
The manifest contains:

- Tested URL
- ISO timestamp
- Telescope package version
- Artifact schema version

The existing performance browser is not reused, and no separate detection browser is launched yet.

### Artifact writer

Adds a reusable `writeBaselineArtifact` function that:

- Namespaces all output beneath `baseline/`.
- Supports nested intermediary paths such as `baseline/detection/`.
- Produces deterministic JSON by sorting object keys.
- Rejects paths that would escape the Baseline directory.
- Accepts only JSON-compatible values through the shared `JsonValue` type.

Later slices can use the same writer for intermediary detection artifacts and the final reports split by Baseline status.

### Error isolation

Baseline pipeline errors are logged without failing the performance test. Existing HAR, metrics, and other performance results remain available if Baseline processing fails.

### Package version helper

Moves the existing package-version lookup from `index.ts` into a focused shared module so both the CLI version flag and Baseline metadata use the same implementation.

## Tests

Adds coverage for:

- `--baseline` appearing in CLI help.
- CLI parsing with the flag enabled and omitted.
- `baseline: true`, `false`, and omitted across the browser matrix using the local delay fixture.
- Creation and contents of `baseline/meta.json`.
- No Baseline directory when the option is disabled or omitted.
- Byte-for-byte isolation of existing HAR and metrics artifacts.
- Deterministic nested artifact output.
- Rejection of paths outside the Baseline directory.
- Pipeline execution after HAR and metrics creation.
- Baseline errors being logged and swallowed without changing the successful performance result.

### DIFF
--- a/packages/telescope/__tests__/baseline.test.ts
+++ b/packages/telescope/__tests__/baseline.test.ts
@@ -1,0 +1,116 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import type { Server } from 'node:http';
+
+import { afterAll, beforeAll, describe, expect, test } from 'vitest';
+
+import { runBaselinePipeline, writeBaselineArtifact } from '../src/baseline.js';
+import { BrowserConfig } from '../src/browsers.js';
+import { launchTest } from '../src/index.js';
+import type { BrowserName } from '../src/types.js';
+import { cleanupTestDirectory } from './helpers.js';
+import {
+  createStaticServer,
+  fixturesDir,
+  listenServer,
+  shutdownServer,
+} from './testServer.js';
+
+const browsers: BrowserName[] = BrowserConfig.getBrowsers();
+const packageVersion = (
+  JSON.parse(fs.readFileSync(path.resolve('package.json'), 'utf8')) as {
+    version: string;
+  }
+).version;
+let baseUrl: string;
+let server: Server;
+
+beforeAll(async () => {
+  server = createStaticServer({ fixturesDirPath: fixturesDir('delay') });
+  baseUrl = await listenServer(server);
+});
+
+afterAll(async () => {
+  await shutdownServer(server);
+});
+
+describe.each(browsers)('Baseline pipeline artifacts (%s)', browser => {
+  test.each([
+    ['enabled', true, true],
+    ['disabled', false, false],
+    ['omitted', undefined, false],
+  ])(
+    'writes artifacts when baseline is %s',
+    async (_case, baseline, enabled) => {
+      const result = await launchTest({
+        baseline,
+        browser,
+        url: `${baseUrl}/index.html`,
+      });
+      if (!result.success) throw new Error(result.error);
+      try {
+        const baselinePath = path.join(result.resultsPath, 'baseline');
+        expect(fs.existsSync(baselinePath)).toBe(enabled);
+        if (enabled) {
+          expect(fs.existsSync(path.join(baselinePath, 'meta.json'))).toBe(
+            true,
+          );
+        }
+      } finally {
+        cleanupTestDirectory(result.testId);
+      }
+    },
+    60000,
+  );
+});
+
+test('the pipeline isolates performance artifacts and writes stable JSON', async () => {
+  fs.mkdirSync(path.resolve('results'), { recursive: true });
+  const resultsPath = fs.mkdtempSync(
+    path.join(process.cwd(), 'results', 'baseline-isolation-'),
+  );
+  const harPath = path.join(resultsPath, 'pageload.har');
+  const metricsPath = path.join(resultsPath, 'metrics.json');
+  const har = '{"log":{"entries":[]}}';
+  const metrics = '{"firstContentfulPaint":123}';
+  fs.writeFileSync(harPath, har);
+  fs.writeFileSync(metricsPath, metrics);
+
+  try {
+    await runBaselinePipeline({
+      resultsPath,
+      url: 'https://example.com',
+    });
+
+    const meta = JSON.parse(
+      fs.readFileSync(path.join(resultsPath, 'baseline', 'meta.json'), 'utf8'),
+    ) as Record<string, unknown>;
+    expect(meta).toEqual({
+      schemaVersion: 1,
+      telescopeVersion: packageVersion,
+      timestamp: expect.any(String),
+      url: 'https://example.com',
+    });
+    expect(new Date(meta.timestamp as string).toISOString()).toBe(
+      meta.timestamp,
+    );
+    expect(fs.readFileSync(harPath, 'utf8')).toBe(har);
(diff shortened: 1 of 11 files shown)


packages/telescope/src/config.ts
```diff
@@ -45,6 +45,7 @@ export function normalizeCLIConfig(options: CLIOptions): LaunchOptions {
   }
   const config: LaunchOptions = {
     url: options.url,
+    baseline: options.baseline || DEFAULT_OPTIONS.baseline,
```
nicjansma: Using `||` here means if the passed-in `options.baseline` is `false`, i.e. `--baseline false`, then it would fallback to `DEFAULT_OPTIONS.baseline` (which is also false at the moment), instead of applying a `false` value.

I think it would be better to use `options.baseline ?? DEFAULT_OPTIONS.baseline` to allow for `false` with a fallback to `DEFAULT_OPTIONS.baseline`

nicjansma (dismissed):
Minor nit, otherwise looks good

packages/telescope/src/config.ts
smsnedden: Good call, I made the change there and then one other place it was showing up as well

packages/telescope/src/types.ts
```diff
@@ -229,6 +231,13 @@ export interface DefaultOptions {
   device?: CustomDeviceDescriptor;
 }
 
+/**
+ * Options used internally after the Baseline default has been applied.
+ */
+export interface TestConfig extends LaunchOptions {
```
sergeychernyshev: I don't understand why we need separate TestConfig here? Why can't we just use LaunchOptions - `baseline` doesn't look any different from any other option.

packages/telescope/src/types.ts
```diff
@@ -643,6 +652,20 @@ export interface HarData {
 // Baseline Types
 // ============================================================================
 
+export type JsonPrimitive = boolean | number | string | null;
```
sergeychernyshev:
```suggestion
export type JSONPrimitive = boolean | number | string | null;
```

nit: Who's Jason? ;)

packages/telescope/src/baseline.ts
```diff
@@ -0,0 +1,67 @@
+import { mkdirSync, writeFileSync } from 'node:fs';
+import path from 'node:path';
+
+import { getPackageVersion } from './packageVersion.js';
+import type { BaselinePipelineOptions, JsonValue } from './types.js';
+
+const BASELINE_SCHEMA_VERSION = 1;
+
+function sortJson(value: JsonValue): JsonValue {
```
sergeychernyshev: Why do we need a custom / deterministic JSON serialization?

If we do, then is there a reason why we should not use it in the whole project when we write JSON rather than just adding it here.

packages/telescope/src/baseline.ts
```diff
@@ -0,0 +1,67 @@
+import { mkdirSync, writeFileSync } from 'node:fs';
+import path from 'node:path';
+
+import { getPackageVersion } from './packageVersion.js';
+import type { BaselinePipelineOptions, JsonValue } from './types.js';
+
+const BASELINE_SCHEMA_VERSION = 1;
```
sergeychernyshev: Should we export this so we can use it in the test?

packages/telescope/__tests__/baseline.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    await runBaselinePipeline({
+      resultsPath,
+      url: 'https://example.com',
+    });
+
+    const meta = JSON.parse(
+      fs.readFileSync(path.join(resultsPath, 'baseline', 'meta.json'), 'utf8'),
+    ) as Record<string, unknown>;
+    expect(meta).toEqual({
+      schemaVersion: 1,
```
sergeychernyshev: Should we use the constant from the file rather than hardcoded `1` value?

packages/telescope/__tests__/baseline.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      z: [{ b: 2, a: 1 }],
+      a: { d: 4, c: 3 },
+    });
+
+    expect(
+      fs.readFileSync(
+        path.join(resultsPath, 'baseline', 'detection', 'example.json'),
+        'utf8',
+      ),
+    ).toBe('{"a":{"c":3,"d":4},"z":[{"a":1,"b":2}]}');
```
sergeychernyshev: Assumes requirement of stable JSON - let's figure out why we need that and document accordingly if needed (see comments on the custom sorting below).

sergeychernyshev (changes requested):
Primary question is for why we need a separate TestConfig type.

The rest are more or less minor changes, but worth resolving for clarity.

packages/telescope/src/types.ts
smsnedden: We actually don't need a separate TestConfig there, I appreciate you pointing that out. I changed it to just use LaunchOptions like you suggested

packages/telescope/src/baseline.ts
smsnedden: My little AI assistant led me a bit astray with that one, I have changed it so that it no longer uses a custom JSON serialization. Thank you for catching that.

packages/telescope/__tests__/baseline.test.ts
smsnedden: Since I switched the approach for sorting based on your comment above, I switched to a round rip (write the artifact, read it back, and `toEqual` the parsed content) so it checks the data persists without depending on key order, and simplified the sample artifact since it no longer needs to show ordering.

sergeychernyshev (approved):
🔭 🚀

Issue #327, opened 2026-07-23
Aggressive page.route disabling HTTP Cache


notzree:
## Describe the bug
From Playwright's own [docs](https://playwright.dev/docs/api/class-page#page-route), calling page.route() disables the HTTP Cache. Specifically for Early Hints, this causes the browser to perform the request again instead of using the hinted resource.
There are two independent causes that we've found:

1. `overrideHost` defaults to `{}`, but the truthy object causes Telescope to
   register a route using an empty regular expression, which matches every URL.

2. Telescope unconditionally registers `page.route('**/*')` to inject a unique
   `x-telescope-id` header into every regular request.

The ID is useful for exact correlation between Playwright request timings, HAR
entries, and Chromium CDP priorities. Removing it silently would break an
existing documented behavior, so a backward-compatible cache-preserving mode
seems preferable.

### Empty `overrideHost` registers a broad route

`overrideHost` defaults to an empty object:

<https://github.com/cloudflare/telescope/blob/9e7f91eb313a20748066daa2ce4aa2bf425e3e72/packages/telescope/src/defaultOptions.ts#L33-L34>

The empty object is truthy, so `setupHostOverrides()` runs:

<https://github.com/cloudflare/telescope/blob/9e7f91eb313a20748066daa2ce4aa2bf425e3e72/packages/telescope/src/testRunner.ts#L307-L308>

With no keys, the implementation evaluates the equivalent of:

```ts
Object.keys({}); // []
[].join("|"); // ''
new RegExp(""); // matches every URL
```

It then passes that expression to `page.route()`:

<https://github.com/cloudflare/telescope/blob/9e7f91eb313a20748066daa2ce4aa2bf425e3e72/packages/telescope/src/testRunner.ts#L138-L149>

Current Telescope also registers a catch-all route to inject
`x-telescope-id`:

<https://github.com/cloudflare/telescope/blob/9e7f91eb313a20748066daa2ce4aa2bf425e3e72/packages/telescope/src/testRunner.ts#L319-L335>

## Telescope info

Please include the following information about your environment:

**Telescope version**: 1.20
**Node version**: v26.0.0
**Operating system version**: Macos Tahoe 26.5
**Google Chrome**: 150.0.7871.129
## Steps to reproduce

1. Confirm that your connection receives the `103`:

   ```bash
   curl --http2 -sS -D - -o /dev/null \
     https://theme-horizon-demo.myshopify.com/
   ```

[...]

PR #340, opened 2026-07-27
Skip page.route() setup when no host overrides configured

sergeychernyshev:
## Summary

The `overrideHost` option defaults to `{}` which is truthy, causing `setupHostOverrides()` to run unconditionally. With no keys, it constructs an empty regex that matches every URL and registers a `page.route()` handler against it. Per [Playwright docs](https://playwright.dev/docs/api/class-page#page-route), calling `page.route()` disables the HTTP Cache — which silently breaks features like Early Hints.

## Changes

- Guard the call site in `preparePage()` to skip `setupHostOverrides()` when `overrideHost` has no entries.
- Add a defensive early-return inside `setupHostOverrides()` itself, so callers that pass an empty object also get a no-op.
- Extend the JSDoc on `setupHostOverrides()` to explain the caching rationale.

## Refs

- Relates to #327 (partial — this addresses only the first root cause; the unconditional `x-telescope-id` catch-all route is still open)

### DIFF
--- a/packages/telescope/src/testRunner.ts
+++ b/packages/telescope/src/testRunner.ts
@@ -159,8 +159,7 @@ class TestRunner {
       const original_domain = original_url.host;
       const override = overrides[original_domain];
       if (!override) {
-        route.fallback();
-        return;
+        return route.fallback();
       }
 
       const new_url = new URL(original_url.toString());
@@ -172,7 +171,7 @@ class TestRunner {
         'X-Host': original_domain,
       };
 
-      route.fallback({ headers, url: new_url.toString() });
+      return route.fallback({ headers, url: new_url.toString() });
     });
 
     return;
@@ -270,7 +269,7 @@ class TestRunner {
       await tmpbrowser.close();
 
       this.selectedBrowser.userAgent = originalUserAgent.concat(
-        " ",
+        ' ',
         this.options.agentExtra.trim(),
       );
     }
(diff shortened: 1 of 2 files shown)


packages/telescope/src/testRunner.ts
```diff
@@ -133,18 +133,27 @@ class TestRunner {
...
+      return;
+    }
 
-    Object.keys(overrides).forEach(original => {
-      domains.push('//(' + original + ')/'); /* Domain part of URL */
-    });
-    const domain_rx = new RegExp(domains.join('|'));
+    const domain_rx = new RegExp(
+      domains.map(original => '//(' + original + ')/').join('|'),
+    );
```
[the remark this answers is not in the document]
sergeychernyshev: Replaced Regex with a matcher function instead

packages/telescope/src/testRunner.ts
```diff
@@ -133,40 +133,46 @@ class TestRunner {
...
+   * Set up any hostname overrides that have been requested.
+   *
+   * Registers a `page.route()` handler only when there is at least one
+   * override entry. Calling `page.route()` disables Playwright's HTTP
+   * cache (see https://playwright.dev/docs/api/class-page#page-route),
+   * so this must be a no-op for the common case where no overrides are
+   * configured — otherwise features like Early Hints break silently.
+   * See https://github.com/cloudflare/telescope/issues/327.
    */
   async setupHostOverrides(
```
MildMax, line 145: Is there a unit test for setupHostOverrides that we could update to test this case hen no overrides are provided?

sergeychernyshev: Added the tests for not calling the page.route() - we have separate ticket for verifying cache-ability which will work after we're done with putting fetchpriority under the flag or finding an alternative to adding headers.

MildMax (approved):

PR #341, opened 2026-07-27
Collect HTML features from the live DOM

smsnedden:
First slice of the `--baseline` HTML analysis feature. This PR adds collection of HTML feature keys from a page's live DOM; collector registration, Baseline-status lookup, and status-grouped output follow in later PRs.

The collector is intentionally **not wired into the CLI or `launchTest`** yet. It has no production caller until the integration PR consumes it, so it is currently exercised directly by its integration tests. Keeping this separate makes the DOM-collection behavior independently reviewable.

### `collectHTMLFeatures(page): Promise<HTMLFeatureCount[]>` (`src/baselineHtmlDetect.ts`)

Collects HTML features from the live main-document DOM:

- Reports recognized elements, element-specific attributes, and global attributes using MDN Browser Compatibility Data keys.
- Handles value-specific features such as `input[type="date"]` and `img[loading="lazy"]`.
- Counts feature occurrences and returns results sorted by BCD key.
- Supports global attributes on custom elements without inventing unsupported element keys.
- Preserves BCD-specific casing such as `html.elements.iframe.privateToken`.
- Ignores unknown attributes, ARIA attributes, roles, event handlers, and invalid bare `data-` attributes.
- Does not modify the document.

The collector intentionally does not traverse shadow roots, child-frame documents, or inert `<template>` contents. Elements outside the HTML namespace are also excluded.

### BCD mapping (`src/baselineHtmlBcd.ts`)

Adds a compact snapshot of HTML element and content-attribute keys from MDN Browser Compatibility Data 8.0.8.

Pinning the mapping keeps results deterministic and allows features to be detected even when the browser running Telescope does not support them. The file documents its source and regeneration process and does not add a runtime dependency.

### Types (`src/types.ts`)

Adds `HTMLFeatureCount`:

```ts
interface HTMLFeatureCount {
  bcdKey: string;
  count: number;
}
```

### Tests
Adds packages/telescope/__tests__/baselineHtmlDetect.test.ts, run across Chromium, Firefox, and WebKit:
- Elements and attributes produce the expected BCD keys and counts.
- Results are sorted deterministically.
- Elements added by JavaScript are included.
- Elements removed by JavaScript are excluded.
- Unknown elements, attributes, and invalid values are ignored.
- Global attributes on custom elements are collected.
- BCD casing aliases and value-specific keys are preserved.
- Collection does not modify the document.


packages/telescope/src/baselineHtmlDetect.ts
```diff
@@ -0,0 +1,168 @@
...
+  'tabindex',
+  'title',
+  'translate',
+  'virtualkeyboardpolicy',
+  'writingsuggestions',
+] as const;
+
+const VALUE_KEYS: Record<string, Record<string, string>> = {
+  'img.loading': {
+    lazy: 'html.elements.img.loading',
```
nicjansma: Is `eager` not listed here because it's the default (legacy) behavior?

nicjansma (dismissed):

packages/telescope/src/baselineHtmlDetect.ts
smsnedden: Yes, exactly. Only `lazy` changes the default behavior

packages/telescope/src/baselineHtmlBcd.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Pinning this data makes detection independent of the tested browser's own
+ * DOM implementation, including for features that browser does not support.
+ *
+ * Source: https://github.com/mdn/browser-compat-data/releases/tag/v8.0.8
+ * To regenerate, read `html.elements` from that release's `data.json`, retain
+ * each element and its first-level keys without underscores, then review the
+ * small set of non-literal entries. DOM-to-BCD casing aliases are documented
+ * beside `ATTRIBUTE_BCD_NAMES` in `baselineHtmlDetect.ts`.
+ */
+export const HTML_ELEMENT_ATTRIBUTES: Readonly<Record<string, string>> = {
```
sergeychernyshev, line 44: Can we instead import this from `@mdn/browser-compat-data` package in some way?

It would be great if we can just upgrade dependency - worst case use a script to build these constants to avoid the drift (@mdn/browser-compat-data already upgraded from v8.0.8 to v8.0.9 as we are reviewing this PR).

packages/telescope/src/baselineHtmlDetect.ts
```diff
@@ -0,0 +1,168 @@
+import type { Page } from 'playwright';
+
+import { HTML_ELEMENT_ATTRIBUTES } from './baselineHtmlBcd.js';
+import type { HTMLFeatureCount } from './types.js';
+
+const GLOBAL_ATTRIBUTES = [
```
sergeychernyshev: Same question of data drift as above.

packages/telescope/src/baselineHtmlDetect.ts
```diff
@@ -0,0 +1,168 @@
...
+      'week',
+    ].map(value => [value, `html.elements.input.type_${value}`]),
+  ),
+};
+
+const VALUE_ONLY_ATTRIBUTES = ['img.loading', 'input.type'] as const;
+
+// HTML lowercases content-attribute names, but this BCD key is camel-cased.
+const ATTRIBUTE_BCD_NAMES: Record<string, string> = {
+  'iframe.privatetoken': 'privateToken',
```
sergeychernyshev: Is that a bug on their end and an inconsistency (why is there just one here)?
Should this be filed as an issue on their repo?

packages/telescope/src/baselineHtmlDetect.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+/**
+ * Collect HTML element and attribute BCD keys from a page's live main-document
+ * DOM. Shadow roots, child-frame documents, and inert `<template>` contents
```
sergeychernyshev: Do we have any description why this is intentional? e.g. phase X or incorrectness? or 3rd party actionability? Might be worth adding documentation for these limitations.

packages/telescope/src/baselineHtmlDetect.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      for (const element of document.querySelectorAll('*')) {
+        if (element.namespaceURI !== 'http://www.w3.org/1999/xhtml') {
+          continue;
+        }
+
+        const tag = element.localName;
+        const supportedAttributes = supportedElements.get(tag);
+        const elementKeys = new Set<string>();
+        if (supportedAttributes) elementKeys.add(`html.elements.${tag}`);
```
sergeychernyshev: Let's always use curly braces.

Can you run the linter `npm run prettier` on the code to see if this is caught by it?

packages/telescope/src/baselineHtmlDetect.ts
```diff
@@ -0,0 +1,168 @@
...
+
+        for (const attribute of element.attributes) {
+          const name = attribute.name.toLowerCase();
+          const value = attribute.value.toLowerCase();
+          let ruleName = `${tag}.${name}`;
+
+          if (name.startsWith('data-') && name.length > 5) {
+            record('html.global_attributes.data_attributes');
+            continue;
```
sergeychernyshev: Should we convert `for` loop with conditional `continue` statements to `filter()` calls instead?

packages/telescope/src/baselineHtmlDetect.ts
sergeychernyshev: In any case, let's move these constants into separate files and import them here.

sergeychernyshev (commented):
It's hacky, but it works.

My main concern is drift and if we can offload maintaining these constants to other packages instead of doing it ourselves.

packages/telescope/src/baselineHtmlBcd.ts
smsnedden: Alright, so after looking into it a bit, `@mdn/browser-compat-data` is about 20MB unpacked, so we probably don't want it as a runtime dependency. There's two main options I'm weighing, the first being a generation script and adding BCD as a devDependency. The other option is to derive from `web-features` instead since we already depend on it, however, this wouldn't include the ~19 deprecated elements we currently have (and that the script would keep), since they don't have a Baseline status. I'm leaning toward the script to keep detection coverage but looking to see what direction you would suggest going forward with this since it would be adding more code and a new dependency.

packages/telescope/src/baselineHtmlDetect.ts
smsnedden: It is an inconsistency in that `privateToken` is the only key that's camel-cased, but the issue doesn't seem to be within MDNs `browser-compat-data`, it's within the spec that it's referencing (https://wicg.github.io/trust-token-api/#iframe-private-token). That being said, I'm not quite sure if an issue should be filed for this.

packages/telescope/src/baselineHtmlDetect.ts
smsnedden: Fair point. `<template>` contents are genuinely not "in use" by the page so this one is a correctness skip. Shadow roots and child-frame documents (same-origin frames only) offer more complete feature coverage, but for the sake of this PR and the time I have left, thinking it's best to leave these out for now. Cross-origin frames we are unable to read their DOM since the browser blocks it, so their features aren't detectable. I can change the wording on the documentation a bit to match this.

packages/telescope/src/baselineHtmlDetect.ts
smsnedden: Sounds good, this is not something caught by `prettier`, it gets caught by ESLint's `curly` rule, though it's not enabled here. Curly braces have now been added here and in three other places it didn't have them.

packages/telescope/src/baselineHtmlDetect.ts
smsnedden: Good question, after looking at both options, I don't think `filter()` fits well here. These loops don't just select a subset, instead they have side effects each iteration (building up `elementKeys` and calling `record(...)`). If the goal here is better readability, I'm happy to pull the pure skips (`role` / `aria-` / `on`) into a small helper (small change, but might read a bit cleaner), just let me know!

smsnedden: This now uses `filter()` and `map()` instead of the `for` loop!

packages/telescope/src/baselineHtmlBcd.ts
smsnedden: Done. The constants are now produced by a generation script using `@mdn/browser-compat-data` as a dev dependency. The generated file is checked in, so BCD is not included as a runtime dependency.

sergeychernyshev (approved):
:rocket:

PR #342, opened 2026-07-27
Baseline status lookup

smsnedden:
This adds a standalone lookup that maps MDN Browser Compatibility Data (BCD) keys to their Baseline status using the `web-features` package.

The HTML, CSS, and JavaScript detectors produce BCD keys, so this provides a shared, detector-independent classification step. Connecting the lookup to those detectors and producing status-grouped output will follow in later changes.

### Baseline status lookup

Adds `lookupBaselineStatus`, which:

- Builds a BCD-key-to-feature-ID index once for constant-time lookups.
- Handles the `feature`, `moved`, and `split` records provided by `web-features`.
- Prefers per-compat-key status when available, falling back to feature-level status.
- Returns `null` for unknown BCD keys.
- Preserves the classification supplied by `web-features`, including `false`, without reclassification.
- Includes feature names, Baseline dates, browser support, and discouraged-feature metadata.
- Returns independent copies of mutable metadata so callers cannot affect subsequent lookups.

### Result types

Adds types describing Baseline classifications, supported browsers, discouraged-feature metadata, and lookup results:

```ts
export interface BaselineLookupResult {
  featureId: string;
  featureName: string;
  baseline: BaselineStatus;
  baselineLowDate: string | null;
  baselineHighDate: string | null;
  support: BaselineSupport;
  discouraged?: Discouraged;
}
```

Baseline date strings are returned as represented by `web-features`, including uncertainty prefixes such as `≤` where applicable.

### Dependency

Adds `web-features` as a runtime dependency. Because it contains evolving Baseline data, keeping classifications current requires periodically updating the resolved version in `package-lock.json` and reviewing any resulting test changes. The compatible version range permits these updates but does not update existing lockfiles automatically. This PR does not introduce an automated refresh schedule.

## Tests

Adds unit coverage for:

- Widely available and Newly available features.
- Unknown BCD keys.
- Preserving an unclassified `false` status.
- Per-compat-key status precedence.
- Discouraged-feature metadata.
- Isolation of returned metadata from caller mutation.
- Omission of discouraged metadata for ordinary features.


packages/telescope/src/types.ts
```diff
@@ -23,15 +23,7 @@ export type Cookie = Parameters<BrowserContext['addCookies']>[0][number];
...
-  | '3g'
-  | '3gfast'
-  | '3gslow'
-  | '2g'
-  | 'cable'
-  | 'dsl'
-  | '4g'
-  | 'fios'
-  | false;
+  '3g' | '3gfast' | '3gslow' | '2g' | 'cable' | 'dsl' | '4g' | 'fios' | false;
```
nicjansma: nit: was this "unwound" to a single line from a formatter, or manually?

(I'd generally prefer each on its own line for readability and maintainability, when an item gets added or removed it just affects a single line).

packages/telescope/src/baselineStatus.ts
```diff
@@ -0,0 +1,86 @@
...
+ * skipped.
+ */
+const bcdKeyToFeatureId = new Map<string, string>();
+for (const [id, feature] of Object.entries(features)) {
+  if (feature.kind !== 'feature') continue;
+  for (const key of feature.compat_features ?? []) {
+    bcdKeyToFeatureId.set(key, id);
+  }
+}
+
```
nicjansma, line 38: Maybe do this only once when `lookupBaselineStatus()` is called, so it's not otherwise always built on import?

nicjansma (changes requested):

packages/telescope/src/types.ts
smsnedden: It was unwound from a formatter, I'll restore the existing one-per-line formatting now.

packages/telescope/src/baselineStatus.ts
smsnedden: Good call, I updated it to build the index on the first lookup and reuse it afterward.

nicjansma (dismissed):

packages/telescope/src/types.ts
sergeychernyshev: Let's use prettier linter to format code and make sure we codify our preferences as rules.

packages/telescope/src/baselineStatus.ts
```diff
@@ -0,0 +1,90 @@
...
+    bcdKeyToFeatureId = new Map<string, string>();
+    for (const [id, feature] of Object.entries(features)) {
+      if (feature.kind !== 'feature') continue;
+      for (const key of feature.compat_features ?? []) {
+        bcdKeyToFeatureId.set(key, id);
+      }
+    }
+  }
+
+  const featureId = bcdKeyToFeatureId.get(bcdKey);
```
sergeychernyshev: Let's define types for these variables so we don't get some weird data as we update the web-features package.

sergeychernyshev (dismissed):
Adding more types would help with maintenance in the future.

packages/telescope/src/baselineStatus.ts
smsnedden: Done. I added explicit types for the feature ID, feature record, and status object.

sergeychernyshev (dismissed):
🚀

sergeychernyshev:
@smsnedden can you resolve this conflict (happened after I merged the other PR) - should be easy.

Then I can approve it and merge it.

smsnedden:
> @smsnedden can you resolve this conflict (happened after I merged the other PR) - should be easy.
> 
> Then I can approve it and merge it.

All done!

sergeychernyshev (approved):

PR #343, opened 2026-07-28
Detect CSS declaration features for Baseline

smsnedden:
Second slice of the `--baseline` CSS analysis feature, following #317. This PR adds detection of CSS declaration features from the previously extracted `CSSSource` objects.

The detector is intentionally **not wired into the CLI or `launchTest`** yet. At-rule occurrence detection, selector detection, Baseline status lookup, and pipeline integration follow in later PRs. Keeping declaration detection separate limits this PR to one independently tested parsing concern.

### `detectCSSFeatures(sources): CSSFeatureDetection[]` (`src/baselineCssDetect.ts`)

Parses CSS with `css-tree` and produces candidate MDN Browser Compatibility Data keys for declaration syntax:

- Detects CSS properties as `css.properties.<property>`.
- Detects single-identifier property values when they can map to property-specific BCD keys.
- Maps CSS-wide keywords such as `revert-layer` through `css.types.global_keywords`.
- Maps custom properties to the generic `css.properties.custom-property` key.
- Maps descriptors through their containing at-rule, including:
  - `@font-face`
  - `@font-palette-values`
  - `@counter-style`
  - `@property`
  - `@function`
  - `@page`
- Distinguishes the `size` and `page-orientation` descriptors from ordinary properties applied within `@page`.
- Normalizes case and escaped CSS identifiers while preserving canonical BCD casing where required.
- Preserves individual occurrences and source lines for later aggregation.
- Reports multiline property values at the line where the value begins.
- Excludes declarations used only as conditional-query operands, such as `@supports (display: grid)`.
- Skips vendor-prefixed declarations and custom names inside `@font-feature-values`.
- Continues processing other stylesheets if one source cannot be parsed; recoverable malformed CSS is handled on a best-effort basis.

Candidate keys are intentionally not validated against `web-features` in this layer. Keys without a corresponding feature will be discarded by the later Baseline lookup stage.

### Types (`src/types.ts`)

Adds a discriminated union for declaration detections:

- `CSSPropertyDetection`
- `CSSPropertyValueDetection`
- `CSSDescriptorDetection`
- `CSSFeatureSource`
- `CSSFeatureDetection`

### Dependencies

Adds `css-tree` for standards-aware CSS parsing, traversal, identifier decoding, and source positions, along with its TypeScript definitions.

## Tests

Adds `packages/telescope/__tests__/baselineCssDetect.test.ts` with 19 pure unit tests covering:

- Property and single-keyword value detection
- Source file and line attribution
- Case normalization and escaped identifiers
- Canonical BCD key casing
- CSS-wide keywords
- Custom and vendor-prefixed properties
- Descriptor namespaces and `@page` classification
- Conditional-query operands
- Repeated feature occurrences
- Malformed stylesheets and empty input

### DIFF
--- a/packages/telescope/src/baselineCssDetect.ts
+++ b/packages/telescope/src/baselineCssDetect.ts
@@ -1,37 +1,18 @@
 import { generate, ident, parse, walk } from 'css-tree';
 
 import type { CssNode, WalkContext } from 'css-tree';
+import {
+  BCD_SEGMENT_PATTERN,
+  CSS_WIDE_KEYWORDS,
+  DESCRIPTOR_AT_RULES,
+  PAGE_DESCRIPTORS,
+  VALUE_CANONICAL_CASE,
+} from './baselineCssBcd.js';
 import type {
   CSSFeatureDetection,
   CSSFeatureSource,
   CSSSource,
 } from './types.js';
-
-// Requiring a leading letter intentionally excludes vendor-prefixed identifiers.
-const BCD_SEGMENT_PATTERN = /^[a-z][a-z0-9-]*$/i;
-const CSS_WIDE_KEYWORDS = new Set([
-  'inherit',
-  'initial',
-  'revert',
-  'revert-layer',
-  'unset',
-]);
-const DESCRIPTOR_AT_RULES = new Set([
-  'counter-style',
-  'font-face',
-  'font-palette-values',
-  'function',
-  'property',
-]);
-const PAGE_DESCRIPTORS = new Set(['page-orientation', 'size']);
-// BCD preserves canonical casing for this small set of CSS keyword segments.
-const VALUE_CANONICAL_CASE = new Map([
-  ['currentcolor', 'currentColor'],
-  ['geometricprecision', 'geometricPrecision'],
-  ['linearrgb', 'linearRGB'],
-  ['nan', 'NaN'],
-  ['srgb', 'sRGB'],
-]);
 
 /**
  * Detect CSS properties and keyword values that can map to MDN Browser
@@ -68,9 +49,13 @@ export function detectCSSFeatures(sources: CSSSource[]): CSSFeatureDetection[] {
           return;
         }
 
-        if (node.type !== 'Declaration' || this.atrulePrelude) return;
+        if (node.type !== 'Declaration' || this.atrulePrelude) {
+          return;
+        }
 
-        if (atRuleStack.includes('font-feature-values')) return;
+        if (atRuleStack.includes('font-feature-values')) {
+          return;
+        }
 
         const location = getSource(node, source.file);
         const decodedProperty = ident.decode(node.property);
@@ -85,7 +70,9 @@ export function detectCSSFeatures(sources: CSSSource[]): CSSFeatureDetection[] {
         }
 
         const property = normalizeName(decodedProperty);
-        if (!property) return;
+        if (!property) {
+          return;
+        }
 
         const parentAtRule = atRuleStack.at(-1);
         const isDescriptor =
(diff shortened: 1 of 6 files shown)


packages/telescope/src/baselineCssDetect.ts
```diff
@@ -0,0 +1,154 @@
...
+
+import type { CssNode, WalkContext } from 'css-tree';
+import type {
+  CSSFeatureDetection,
+  CSSFeatureSource,
+  CSSSource,
+} from './types.js';
+
+const CSS_NAME_PATTERN = /^[a-z][a-z0-9-]*$/i;
+const CSS_VALUE_PATTERN = /^[a-z][a-z0-9-]*$/i;
```
nicjansma: Same regexs, worthwhile to collapse to a single variable?

For `CSS_NAME_PATTERN`, this would exclude vendor-prefixed names `-ms-xyz`, maybe good to call that out if we keep the two separate.

nicjansma (changes requested):

packages/telescope/src/baselineCssDetect.ts
smsnedden: Definitely worthwhile to collapse to a single variable, that change should be made now.

nicjansma (dismissed):

packages/telescope/src/baselineCssDetect.ts
```diff
@@ -0,0 +1,154 @@
...
+import type { CssNode, WalkContext } from 'css-tree';
+import type {
+  CSSFeatureDetection,
+  CSSFeatureSource,
+  CSSSource,
+} from './types.js';
+
+// Requiring a leading letter intentionally excludes vendor-prefixed identifiers.
+const BCD_SEGMENT_PATTERN = /^[a-z][a-z0-9-]*$/i;
+const CSS_WIDE_KEYWORDS = new Set([
```
sergeychernyshev: Let's organize constants into a separate file for maintenance or import them if possible.

Also, with amount of baseline files, let's move them into subfolder, probably as separate PR, no need to do it here.

packages/telescope/src/baselineCssDetect.ts
```diff
@@ -0,0 +1,154 @@
...
+
+        detections.push({
+          type: 'property',
+          bcdKey: `css.properties.${property}`,
+          property,
+          source: location,
+        });
+
+        const value = normalizeValue(generate(node.value));
+        if (!value) return;
```
sergeychernyshev: Looks like prettier actually allows these abominations (if blocks without curly braces) as long as they are on one line, e.g. short.

I am personally not a fan, but can live with that, in any case, it'll require a plugin config so ignore all my comments about them, but run prettier anyway to format other issues.

sergeychernyshev (dismissed):
Ideally, we should move out all constants listing all the web-features which we use for massaging the data structures out so we can figure out how to better maintain them.

It can also wait, but old Russian adage of "there is nothing more permanent than something temporary" tells me that the sooner we do that, the less pain it'll bring us.

packages/telescope/src/baselineCssDetect.ts
smsnedden: I was making some changes anyways so decided to add those curly braces in anyways to keep it consisten. I've made a note to include those in the future as well.

sergeychernyshev (approved):
<s>Let's go!</s> not yet ;)

packages/telescope/src/types.ts
```diff
@@ -645,6 +645,42 @@ export interface HarData {
...
+  source: CSSFeatureSource;
+}
+
+/** A CSS property-value pair mapped to its candidate BCD key. */
+export interface CSSPropertyValueDetection {
+  type: 'property-value';
+  bcdKey: string;
+  property: string;
+  source: CSSFeatureSource;
+  value: string;
```
sergeychernyshev, line 683: Looks like something got cut off here - typescript is throwing errors during build

sergeychernyshev (changes requested):

packages/telescope/src/types.ts
smsnedden: Sure enough, should be all fixed now. Thank you

sergeychernyshev (approved):
🚀

PR #348, opened 2026-08-10
Add production-safe Baseline detection pass

smsnedden:
Adds the standalone browser pass that will support live-DOM Baseline detection after the performance run completes.

The pass is intentionally **not wired into the CLI or `launchTest` yet**. Pipeline integration, combining HAR and live CSS sources, and writing detection artifacts follow in a separate PR. Keeping the browser lifecycle separate makes its request-context behavior, failure isolation, and cleanup independently reviewable.

### Baseline detection pass (`src/baselineDetectionPass.ts`)

Adds `runBaselineDetectionPass`, which:

- Launches a throwaway browser using the selected Playwright engine and browser configuration.
- Recreates content-affecting context settings:
  - Authentication
  - Headers
  - Cookies
  - Viewport and device scale
  - User agent
  - Mobile and touch settings
  - JavaScript enablement
- Clones cookies before applying URL defaults rather than mutating caller-owned input.
- Accepts a page-preparation callback for content-affecting routing such as blocking and host overrides.
- Navigates once using `networkidle`.
- Collects inline `<style>` sources from the live DOM through the existing `harvestInlineStyles` collector.

HAR parsing, source consolidation, artifact writing, and pipeline invocation remain out of scope for this PR.

### Timeout and failure handling

- Uses Telescope’s configured navigation timeout, falling back to the canonical default for non-positive values.
- Recognizes navigation timeouts by Playwright error name or timeout message, matching Telescope’s existing navigation behavior.
- Stops network activity after a timeout before collecting partial results.
- Propagates unexpected navigation and setup failures.
- Isolates collector failures so one collector does not discard other detection results.
- Always attempts to close the browser context and browser.
- Prevents cleanup failures from replacing successful results or the original failure.

### Safe diagnostics (`src/baselineError.ts`)

Adds a reusable Baseline error formatter that:

- Handles `Error` objects and unknown thrown values.
- Redacts HTTP and HTTPS URLs that may contain credentials or tokens.
- Safely handles values that cannot be converted to strings.

### Types (`src/types.ts`)

Adds:

- `BaselineDetectionPassOptions`
- `BaselineDetectionPassResult`

These define the browser configuration, request context, page preparation, timeout, and collected live-DOM sources used by the shared pass.

## Tests

Adds direct real-browser and mocked lifecycle coverage for:

- Browser launch and context configuration forwarding.
- Headers, cookies, authentication, and page preparation.
- Request blocking before navigation.
- Live inline CSS collection.
- Named and message-based navigation timeouts.
- Default timeout fallback.
- Partial collection after timeout.
- Non-timeout navigation failures.
- Context-creation and page-preparation failures.
- Context and browser cleanup failures.
- Preservation of successful results when cleanup fails.
- Collector failure isolation.
- URL redaction and malformed thrown values.
- Cookie URL defaults, domain cookies, arrays, and input immutability.

The direct browser integration test is required while the pass has no production caller. A follow-up pipeline PR may replace it with equivalent end-to-end coverage only after preserving its header, cookie, blocking, and browser behavior assertions.


packages/telescope/src/types.ts
```diff
@@ -678,6 +678,20 @@ export interface CSSSource {
...
+  cookies?: Cookie | Cookie[];
+  headers?: Record<string, string>;
+  preparePage: (page: Page) => Promise<void>;
+  timeout: number;
+  url: string;
+}
+
+/** Sources collected from the live page during the shared detection pass. */
+export interface BaselineDetectionPassResult {
+  inlineCSSSources: CSSSource[];
```
sergeychernyshev, line 732: Another cutoff in the types, try running the build.

sergeychernyshev (changes requested):

packages/telescope/src/types.ts
smsnedden: Ah, I don't know why it keeps doing that, all fixed now, I ran the build as well.

sergeychernyshev (dismissed):

sergeychernyshev (approved):

MildMax (approved):

sufian-cf (approved):

PR #349, opened 2026-08-10
Baseline JavaScript API availability

smsnedden:
Adds a standalone registry and collector for determining which curated JavaScript and Web APIs are available in a browser.

The collector is intentionally **not wired into the CLI, `launchTest`, or the Baseline detection pass yet**. Detection-pass registration and artifact generation follow separately. Keeping this slice standalone makes the registry contract and browser inspection behavior independently reviewable.

This reports APIs exposed by the detection browser, not APIs used by the tested page. Runtime usage instrumentation follows separately.

### JavaScript API registry (`src/baselineJsApiRegistry.ts`)

Adds a curated registry of 98 JavaScript and Web APIs selected for reliable runtime detection.

Each entry records:

- The corresponding MDN Browser Compatibility Data key.
- Its path from the page's `globalThis` object.
- Its exposure kind: constructor, global function, method, property, or static method.

The registry does not mirror the complete BCD dataset. It focuses on commonly used APIs that can be detected safely and reliably at runtime. BCD keys identify the exact exposure being checked rather than only the interface type returned by that exposure.

Because the registry is manually curated, existing keys should be validated whenever the compatibility dataset is updated. New entries should be considered as commonly used APIs emerge or detection coverage is expanded. This PR does not introduce an automated refresh process.

### Availability collector (`src/baselineJsApiAvailability.ts`)

Adds `collectAvailableJSAPIs`, which inspects an already-loaded Playwright page and returns the BCD keys for available registry entries.

The collector:

- Checks each API independently so a missing or throwing browser property does not prevent other APIs from being collected.
- Uses property existence for property entries without invoking the final getter.
- Requires constructors, functions, and methods to be callable.
- Does not replace, wrap, or otherwise modify browser APIs.
- Deduplicates and sorts results by BCD key for deterministic output.
- Accepts an alternate registry for focused testing and future reuse.

Browser launch, navigation, request-context setup, usage instrumentation, and artifact writing remain out of scope for this PR.

### Types (`src/types.ts`)

Adds:

- `JSAPIKind`
- `JSAPIRegistryEntry`

These define the shared contract used by the registry and collector.

## Tests

Adds `packages/telescope/__tests__/baselineJsApiAvailability.test.ts` with registry validation and real-browser coverage across Chromium, Firefox, and WebKit:

- Representative constructor, function, method, property, and static-method entries.
- Non-empty and unique BCD keys and global paths.
- Available, missing, and non-callable APIs.
- Isolation of intermediate getters that throw.
- Property availability checks without invoking the final getter.
- Sorted and deduplicated results.
- Preservation of existing browser API identities.
- Detection of representative entries from the production registry.

All 98 registry keys were also checked against the current `web-features` dataset.


sergeychernyshev (dismissed):
Just need to resolve some conflicts.

sergeychernyshev (dismissed):
:rocket:

sergeychernyshev (dismissed):

MildMax (dismissed):

sufian-cf (dismissed):

sufian-cf (approved):

PR #351, opened 2026-08-14
Add test id to the tmp directory file path

icrutche:
Currently the tmp directory created to store browser context will always use `./tmp`, meaning multiple runs of telescope will share it, causing test failures due to the browser not allowing shared directories. Adding the test id means each run will use a separate directory. The directory is deleted on completion of the test, so this won't increase disk usage.


sergeychernyshev (approved):
Perfect! I had a nagging feeling that I missed this detail from the past browsing through code.

